PR #1, opened 2024-03-19
Add draft layout of the app

Diogomartf:
This PR is meant to draft the main layout of the app.

**Landing**
image

**My bets**
image

**Question page**
image


next.config.mjs
```diff
@@ -1,4 +1,6 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {};
+const nextConfig = {
+  distDir: "dist",
+};
```
berteotti, line 4: Let's add this too, so we can export the app to IPFS and fix webpack lib issues.

```suggestion
const nextConfig = {
  distDir: "dist",
  output: "export",
  webpack(config) {
    config.resolve.fallback = { "pino-pretty": false, encoding: false };

    return config;
  },
};
```

app/my/page.tsx
```diff
@@ -0,0 +1,19 @@
+import { RowCard } from "@/app/components/ui";
+import Link from "next/link";
+
+export default function MyPage() {
```
berteotti, line 4: "My" is a bit vague, my-bets would be better. Other option is to call it portfolio.

Diogomartf: ok, we can change it to my bets

berteotti: We will show only bets there, so I think my-bets works fine for now. Or are you going to show more things under `/my/`?

Diogomartf: just bets

berteotti (approved):

PR #6, opened 2024-04-03
Add CardMarket component and add font

Diogomartf:
- Add `CardMaket` component (mock data for now)
- Add font Manrope

image


utils/dates.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (count >= 1) {
+      if (past) {
+        return `Happened ${count} ${unit.label}${count !== 1 ? "s" : ""} ago`;
+      } else {
+        return `${count} ${unit.label}${count !== 1 ? "s" : ""} remaining`;
+      }
+    }
+  }
+
+  return "Less than a second remaining";
```
berteotti, line 26: Let's use `date-fns` here for this logic

app/components/CardMarket.tsx
```diff
@@ -0,0 +1,74 @@
...
+        <div className="flex space-x-4 ">
+          <div className="size-[40px] bg-text-low-em rounded-8 bg-gradient-to-r from-[#cb8fc1] to-[#b459c6]" />
+          <ScrollArea className="flex-1 font-semibold text-text-high-em h-[98px]">
+            {market.title}
+          </ScrollArea>
+        </div>
+        <OutcomeBar
+          market={market}
+          outcome0Price={outcome0Price}
+          outcome1Price={outcome1Price}
```
berteotti, line 38: Let's pass only `market` prop here, we can calculate `outcomeXPrice` inside the component

app/components/OutcomeBar.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export const OutcomeBar = ({
+  market,
+  outcome0Price,
+  outcome1Price,
+}: OutcomeBarProps) => {
+  function formatRoundNumber(number: string | number) {
+    number = parseFloat(number.toString()); // Parse the number to handle string inputs as well
+    return Math.ceil(number); // Round up to the nearest integer
+  }
```
berteotti, line 20: we can just used to `.toFixed(2)` I guess

app/components/OutcomeBar.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <div
+        className="flex items-center h-8 px-2 bg-surface-success-accent-1 rounded-s-8"
+        style={{ width: `${option0.roundedPrice}%` }}
+      >
+        <p className="w-full uppercase text-text-success-em">
+          {`${option0.title} - ${option0.price}%`}
+        </p>
+      </div>
+      <div
+        className="flex items-center w-full h-8 px-2 bg-surface-danger-accent-1 rounded-e-8"
```
berteotti, line 47:
```suggestion
        className="flex items-center h-8 px-2 bg-surface-danger-accent-1 rounded-e-8"
```

No need for width styling.

app/components/OutcomeBar.tsx
```diff
@@ -0,0 +1,56 @@
...
+    price: outcome0Price,
+  };
+  const option1 = {
+    title: market.outcomes?.[1],
+    roundedPrice: formatRoundNumber(outcome1Price),
+    price: outcome0Price,
+  };
+
+  if (option0.roundedPrice <= 0 && option1.roundedPrice <= 0)
+    return <Tag colorScheme="quaternary">Closed</Tag>;
```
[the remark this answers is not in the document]
Diogomartf: yeah, we have to improve it, but showing slash was bothering me, so let's leave it like this for now and update it later.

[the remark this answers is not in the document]
berteotti: I would rather show slashes than making unthoughtful decisions. This is showing that a market is closed when it's open and accepting bets. I don't know what "no option price" means at this time, but the phrasing should be aligned what the meaning. "Closed" isn't the same as "no option price".

With slashes, the problem is evident. With "Closed" Tag it looks like it's solved (or just missing some better styling).

utils/dates.ts
```diff
@@ -0,0 +1,13 @@
+import { formatDistance } from "date-fns";
+
+export const remainingTime = (date: Date): string => {
+  const now: Date = new Date();
+  const timeDiff: number = new Date(date).getTime() - now.getTime();
+  const isInThePast: boolean = timeDiff < 0;
```
berteotti, line 6: We can also use `isPast` from `date-fns`

```suggestion
  const isInThePast = isPast(date)
```

berteotti (approved):

PR #7, opened 2024-04-04
feat: create swapbox header

berteotti:
image


app/markets/[id]/MarketDetails.tsx
```diff
@@ -0,0 +1,97 @@
...
+  if (error) throw error;
+  if (!data?.fixedProductMarketMaker) return <LoadingMarketDetails />;
+
+  const market = data.fixedProductMarketMaker;
+
+  const closingDate = new Date(+market.openingTimestamp * 1000);
+
+  return (
+    <div className="space-y-4">
+      <Link className="flex space-x-1.5 items-center w-fit" href="/">
```
Comment (Diogomartf) [inline app/markets/[id]/MarketDetails.tsx:29]: This section should do hover together

```suggestion
      <Link className="flex space-x-1.5 items-center w-fit group" href="/">
```

Diogomartf (approved):
Good job Leo 🥳

app/markets/[id]/MarketDetails.tsx
Comment (Diogomartf) [reply app/markets/[id]/MarketDetails.tsx:29]: then we need to add `group-hover:` modifier bellow.

https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state

But as they are imported components not sure if it will take to much work. This is a nice to have.

PR #8, opened 2024-04-04
Add CardBet

Diogomartf:
- Add loading with LoadingCardBet
- data is just for mocking

image


berteotti (commented):
Looking good, only missing a good mobile view

image

berteotti:
Also, should we have gradients already at this stage?

Diogomartf:
- Will improve mobile.
- I can add the gradients as well.

app/components/CardBet.tsx
```diff
@@ -0,0 +1,92 @@
...
+    <Card className="w-full">
+      <Link key={market.id} href={`markets/${market.id}`} className="block">
+        <section className="p-4 h-[144px] flex flex-col justify-between space-y-4">
+          <div className="flex justify-between items-center">
+            <div className="flex space-x-2">
+              <Tag colorScheme="quaternary" size="sm">
+                Election
+              </Tag>
+              <Tag colorScheme="success" size="sm">
+                You choose Yes
```
berteotti, line 26: Just to have a good example, not necessary at this stage but I prefer to have it right from the start.
We should use the past tense here.

```suggestion
                You chose Yes
```

app/my-bets/page.tsx
```diff
@@ -1,17 +1,75 @@
...
+              <TabStyled>
+                All bets{" "}
+                <div className="bg-surface-surface-0 text-2xs border border-outline-low-em rounded-6 p-1 px-1.5 ml-2">
+                  09
+                </div>
+              </TabStyled>
+              <TabStyled>Active</TabStyled>
+              <TabStyled>Unredeemed</TabStyled>
+              <TabStyled>Complete</TabStyled>
+            </TabHeader>
```
berteotti, line 44: This should be scrollable on mobile. Otherwise it will break the window size.

berteotti (approved):

PR #11, opened 2024-04-10
Add Swapbox footer

Diogomartf:
- add default html styles (missing now that we removed `swapr-ui/styles`)

image


app/markets/[id]/MarketDetails.tsx
```diff
@@ -78,21 +78,30 @@ const LoadingMarketDetails = () => (
...
         <div className="flex space-x-4">
           <div className="size-20 rounded-8 flex-shrink-0 bg-outline-low-em animate-pulse" />
-          <div className="w-full h-20 rounded-8 bg-outline-low-em animate-pulse" />
+          <div className="w-full h-28 rounded-8 bg-outline-low-em animate-pulse" />
         </div>
-        <div className="!mt-7 w-full h-7 rounded-8 bg-outline-low-em animate-pulse" />
+        <div className="!mt-7 w-full h-10 rounded-8 bg-outline-low-em animate-pulse" />
       </div>
-      <div className="p-2 w-full h-[348px]">
+      <div className="p-2 w-full h-[112px]">
```
Comment (ElRodrigote) [inline app/markets/[id]/MarketDetails.tsx:93]:
```suggestion
      <div className="p-2 w-full h-28">
```

app/markets/[id]/MarketDetails.tsx
```diff
@@ -78,21 +78,30 @@ const LoadingMarketDetails = () => (
...
         <div className="w-full h-full rounded-8 bg-outline-low-em animate-pulse" />
       </div>
-    </div>
+      <div className="p-2 w-full h-[108px]">
+        <div className="w-full h-full rounded-8 bg-outline-low-em animate-pulse" />
+      </div>
+      <div className="p-2 w-full h-[54px]">
+        <div className="w-full h-full" />
+      </div>
+      <div className="p-2 w-full h-[64px]">
```
Comment (ElRodrigote) [inline app/markets/[id]/MarketDetails.tsx:102]:
```suggestion
      <div className="p-2 w-full h-16">
```

app/markets/[id]/MarketDetails.tsx
Comment (berteotti) [reply app/markets/[id]/MarketDetails.tsx:93]: h-28 doesn't correspond to 112px since our base font-size is 14px and sizes are calculated in rem. Default size is 16px, so h-28 will be a little less than 112px. I'm ok with either since this is a loading skeleton.

app/components/Swapbox.tsx
```diff
@@ -36,10 +36,30 @@ export const Swapbox = () => {
       <SwapInput
         title="To Receive"
         value={tokenOutAmount}
-        onChange={(event) => {
+        onChange={event => {
           setTokenOutAmount(event.target.value);
         }}
       />
+      <section>
```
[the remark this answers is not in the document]
Diogomartf: I actually like to use `section` to divide things that are their own section. 
Because I see it as a `div`, with the same difference. Because it's handy to have that visual distinction.
But yeah, reading the docs, it should have a heading title to be a section.

Do you guys know an element that could do this?
 
**Section docs**
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section

app/markets/[id]/MarketDetails.tsx
Comment (Diogomartf) [reply app/markets/[id]/MarketDetails.tsx:93]: Yeah, I think we can stay with the `px`.

But changing the base font breaks a lot of tailwind conventions, we probably should be using the normal base 16px and set the default size on the `body` tag as `text-sm`.

what do you think?

Comment (berteotti) [reply app/markets/[id]/MarketDetails.tsx:93]: That sounds better, let's do that.

Comment (Diogomartf) [reply app/markets/[id]/MarketDetails.tsx:93]: Created a task. I'll be fixed.

app/components/Swapbox.tsx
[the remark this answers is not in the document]
Diogomartf: I think that's too much. I don't think I want more than `html` at this point. Not a big deal, it's just nice to have that visual division.

berteotti (approved):

PR #13, opened 2024-04-11
feat: add betting logic

berteotti:
- Add betting logic to Swapbox
- Adds query to filter account's markets. Task: PGIO-10

image

### DIFF
--- a/app/components/Swapbox.tsx
+++ b/app/components/Swapbox.tsx
@@ -2,30 +2,155 @@
 
 import { Button, IconButton } from "swapr-ui";
 import { SwapInput } from "./ui/SwapInput";
-import { useState } from "react";
+import { useEffect, useState } from "react";
+import { erc20Abi, formatEther, parseEther, Address } from "viem";
+import { useAccount, useConfig, useReadContract } from "wagmi";
+import { waitForTransactionReceipt, writeContract } from "wagmi/actions";
+import MarketABI from "@/abi/market.json";
+import { ConnectButton } from ".";
+import { useReadCalcBuyAmount } from "@/model/market";
 
-export const Swapbox = () => {
+const WXDAI_ADDRESS = "0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d";
+const SLIPPAGE = 0.01;
+const ONE_SHARE = "1";
+
+// Removes the given fraction from the given integer-bounded amount and returns the value as an original type.
+const removeFraction = (amount: bigint, fraction: number): bigint => {
+  if (fraction >= 1 || fraction <= 0)
+    throw `The given basisPoints ${fraction} is not in the range [0, 1].`;
+
+  const basisPoints = 10000;
+
+  const fractionInBasisPoints = fraction * basisPoints;
+
+  const keepFraction = basisPoints - fractionInBasisPoints;
+
+  return (amount * BigInt(keepFraction)) / BigInt(basisPoints);
+};
+
+export const Swapbox = ({ id }: { id: Address }) => {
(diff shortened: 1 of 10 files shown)


app/components/Swapbox.tsx
```diff
@@ -2,11 +2,117 @@
...
+  const { address } = useAccount();
+  const config = useConfig();
 
-export const Swapbox = () => {
   const [tokenInAmount, setTokenInAmount] = useState("");
   const [tokenOutAmount, setTokenOutAmount] = useState("");
+  const [minAmountOut, setMinAmountOut] = useState<bigint>();
+
+  const outcomeIndex = 0;
+  const slippage = 0.01;
```
Diogomartf, line 36: this should be a constant right? All caps.

app/components/Swapbox.tsx
```diff
@@ -2,11 +2,117 @@
...
+  const amountWei = parseEther(tokenInAmount);
+
+  const { data: buyAmount } = useReadContract({
+    abi: MarketABI,
+    address: id,
+    functionName: "calcBuyAmount",
+    args: [amountWei, outcomeIndex],
+    query: { enabled: !!tokenInAmount },
+  });
+
```
Diogomartf, line 47: just an idea to explor, if we want to extract this into a file,

this could be on eg. `models/market` and be a function like.

`getBuyAmount({id, tokenInAmount})`

`amountWei` could be calculated inside..

Same as the ones bellow.

app/markets/[id]/MarketDetails.tsx
```diff
@@ -6,9 +6,10 @@ import { getMarket } from "@/queries/omen";
 import { Button, IconButton, Tag } from "swapr-ui";
 import { remainingTime } from "@/utils/dates";
 import Link from "next/link";
+import { Address } from "viem";
 
 interface MarketDetailsProps {
-  id: string;
+  id: Address;
 }
```
Comment (Diogomartf) [inline app/markets/[id]/MarketDetails.tsx:13]: why not call id, address?

Comment (berteotti) [reply app/markets/[id]/MarketDetails.tsx:13]: We can call it address, but I prefer to call id as it is the market identifier which is an address.

app/components/Swapbox.tsx
berteotti: I like to have this logic next to where it's being used. But I think we can abstract some things, like creating a  `useReadMarketContract` hook. If needed we can even abstract more things in the future.

app/components/Swapbox.tsx
berteotti: We can have that. But since it's going to be a variable value I used lowercase. Will update `slippage` but `outcomeIndex` will be a variable value soon.

app/components/Swapbox.tsx
```diff
@@ -2,30 +2,155 @@
...
+const WXDAI_ADDRESS = "0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d";
+const SLIPPAGE = 0.01;
+const ONE_SHARE = "1";
+
+// Removes the given fraction from the given integer-bounded amount and returns the value as an original type.
+const removeFraction = (amount: bigint, fraction: number): bigint => {
+  if (fraction >= 1 || fraction <= 0)
+    throw `The given basisPoints ${fraction} is not in the range [0, 1].`;
+
+  const basisPoints = 10000;
```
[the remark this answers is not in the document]
berteotti: One basis point is equivalent to one one-hundredth of one percent. We are using this to handle division with `bigint` in JS. There is no way to multiply a `bigint` with a float, so we need to turn the float percentage into basis points percentage, so we can use it as a `bigint`.

PR #15, opened 2024-04-19
feat: add sell outcome logic

berteotti:
Adds functions and logic to compute the sell amount necessary with user provided outcome token amount.

image


model/conditionalTokens.ts
```diff
@@ -0,0 +1,68 @@
+import { UseReadContractParameters, useReadContract } from "wagmi";
+import ConditionalTokensABI from "@/abi/conditionalTokens.json";
+import { Abi, Address, zeroHash } from "viem";
+
+export const CONDITIONAL_TOKEN_ADDRESS =
```
Diogomartf, line 5:
```suggestion
export const CONDITIONAL_TOKEN_CONTRACT_ADDRESS =
```

model/conditionalTokens.ts
```diff
@@ -0,0 +1,68 @@
...
+  return useReadContract({
+    abi: ConditionalTokensABI as Abi,
+    address: CONDITIONAL_TOKEN_ADDRESS,
+    functionName,
+    args,
+    query,
+  });
+};
+
+export const useReadBalanceOf = (
```
Diogomartf, line 22: I know we are using the `useRead` naming to imply we are using useReadContract under the hood.
But I'm wondering, can we get that assumption, and replace `useRead` with `get`?

```suggestion
export const getBalanceOf = (
```

`get` is straightforward, `useRead` give us insight in the mechanics.

Diogomartf (approved):

model/conditionalTokens.ts
berteotti: We are using hooks we need to use `use` keyword. Since we need to use this keyword, we can use `useGet` or `useRead`. I think using the latter is better since we are already using wagmi `useRead`.

PR #16, opened 2024-04-19
My bets data

Diogomartf:
**summary**
- fetch market data for position on BetCard 
- add queries for conditionals tokens subgraph
- fetch data from conditionals token subgraph
- update my bets page to fetch userPositions data
- add subgraph urls from constants

image

Out of scope:
- outcome amount calculations
- correct answer

### DIFF
--- a/app/components/CardBet.tsx
+++ b/app/components/CardBet.tsx
@@ -8,6 +8,7 @@ import { useQuery } from "@tanstack/react-query";
 import { cx } from "class-variance-authority";
 import Link from "next/link";
 import { Button, Logo, Tag } from "swapr-ui";
+import { fromHex } from "viem";
 
 interface BetProps {
   userPosition: UserPosition;
@@ -16,7 +17,7 @@ interface BetProps {
 export const CardBet = ({ userPosition }: BetProps) => {
   const conditionId = userPosition.position.conditionIdsStr;
 
-  const { data, isLoading, isFetching } = useQuery({
+  const { data, isLoading } = useQuery({
     queryKey: ["getConditionMarket", conditionId],
     queryFn: async () =>
       getConditionMarket({
@@ -31,18 +32,20 @@ export const CardBet = ({ userPosition }: BetProps) => {
     ?.fixedProductMarketMakers[0] as FixedProductMarketMaker;
 
   const closingDate = new Date(+market.openingTimestamp * 1000);
-  const answer = market.question?.currentAnswer;
-
+  const answer =
+    market.question.currentAnswer &&
+    fromHex(market.question.currentAnswer, "number");
   const outcomeIndex = userPosition.position.indexSets[0];
 
   const outcomes = userPosition.position.conditions[0].outcomes;
   const outcomeString = outcomes ? outcomes[outcomeIndex - 1] : "";
 
-  const winner = answer === "win";
-  const loser = answer === "loss";
+  const hasFinished = !!answer;
+  const isWinner = hasFinished && answer == outcomeIndex;
+  const isLoser = hasFinished && answer != outcomeIndex;
 
   const outcomeAmountString = answer
-    ? winner
+    ? isWinner
       ? "You won"
       : "You lost"
     : "Potential win";
@@ -51,9 +54,9 @@ export const CardBet = ({ userPosition }: BetProps) => {
     <Card
       className={cx(
         "w-full bg-gradient-to-b from-[#F1F1F1] dark:from-[#131313]",
-        winner &&
+        isWinner &&
           "from-[#F2f2F2] to-[#d0ffd6] dark:from-[#131313] dark:to-[#11301F]",
-        loser &&
+        isLoser &&
           "from-[#F2f2F2] to-[#f4cbc4] dark:from-[#131313] dark:to-[#301111]"
       )}
     >
@@ -110,7 +113,7 @@ export const CardBet = ({ userPosition }: BetProps) => {
               />
             </div>
           </div>
-          {answer && (
+          {isWinner && (
             <Button size="sm" colorScheme="success" variant="pastel">
               Reedem
             </Button>
--- a/app/components/CardBet.tsx
+++ b/app/components/CardBet.tsx
@@ -1,39 +1,71 @@
 "use client";
 
 import { Card } from "@/app/components/ui";
-import { FixedProductMarketMaker } from "@/queries/omen";
+import { UserPosition } from "@/queries/conditional-tokens/types";
+import { FixedProductMarketMaker, getConditionMarket } from "@/queries/omen";
 import { remainingTime } from "@/utils/dates";
+import { useQuery } from "@tanstack/react-query";
 import { cx } from "class-variance-authority";
 import Link from "next/link";
 import { Button, Logo, Tag } from "swapr-ui";
 
-interface CardBetProps {
-  market: FixedProductMarketMaker;
+interface BetProps {
+  userPosition: UserPosition;
 }
 
-export const CardBet = ({ market }: CardBetProps) => {
-  const closingDate = new Date(+market.openingTimestamp * 1000);
+export const CardBet = ({ userPosition }: BetProps) => {
+  const conditionId = userPosition.position.conditionIdsStr;
 
-  const outcomes = ["win", "loss", "-"];
-  const randomOutcome = outcomes[Math.floor(Math.random() * outcomes.length)];
+  const { data, isLoading, isFetching } = useQuery({
+    queryKey: ["getConditionMarket", conditionId],
+    queryFn: async () =>
+      getConditionMarket({
+        id: conditionId,
+      }),
+    enabled: !!conditionId,
+  });
+
+  if (isLoading || isFetching) return <LoadingCardBet />;
+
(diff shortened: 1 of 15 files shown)


models/omen/index.ts
berteotti, line 1: We can delete this, right?

Diogomartf: yes!

app/components/CardBet.tsx
```diff
@@ -1,39 +1,71 @@
...
+    enabled: !!conditionId,
+  });
+
+  if (isLoading || isFetching) return <LoadingCardBet />;
+
+  const market: FixedProductMarketMaker = data?.conditions[0]
+    ?.fixedProductMarketMakers[0] as FixedProductMarketMaker;
+
   const closingDate = new Date(+market.openingTimestamp * 1000);
+  const answer = market.question?.currentAnswer;
```
berteotti, line 34: This value need to be converted into an integer. Use `fromHex` from viem to do this.

```suggestion
  const answer = fromHex(
      market.question?.currentAnswer,
      "number"
    );
```

Answer will the winner index. Which we can compare to `outcomeIndex - 1`

app/components/CardBet.tsx
```diff
@@ -1,39 +1,71 @@
...
+  const outcomeString = outcomes ? outcomes[outcomeIndex - 1] : "";
+
+  const winner = answer === "win";
+  const loser = answer === "loss";
+
+  const outcomeAmountString = answer
+    ? winner
+      ? "You won"
+      : "You lost"
+    : "Potential win";
```
berteotti, line 48: Let's have just a `isWinner` boolean.

berteotti: Or `hasWon`

app/components/CardBet.tsx
```diff
@@ -1,39 +1,71 @@
...
     <Card
       className={cx(
-        "w-full bg-gradient-to-b from-[#131313]",
-        randomOutcome === "win" && "from-[#131313] to-[#11301F]",
-        randomOutcome === "loss" && "from-[#131313] to-[#301111]"
+        "w-full bg-gradient-to-b from-[#F1F1F1] dark:from-[#131313]",
+        winner &&
+          "from-[#F2f2F2] to-[#d0ffd6] dark:from-[#131313] dark:to-[#11301F]",
+        loser &&
+          "from-[#F2f2F2] to-[#f4cbc4] dark:from-[#131313] dark:to-[#301111]"
```
berteotti, line 57: Same here here. If it `hasFinished` and `isWinner` show one gradient, otherwise show the other.

app/components/CardBet.tsx
```diff
@@ -1,39 +1,74 @@
...
+    market.question.currentAnswer &&
+    fromHex(market.question.currentAnswer, "number");
+  const outcomeIndex = userPosition.position.indexSets[0];
 
-  const outcomes = ["win", "loss", "-"];
-  const randomOutcome = outcomes[Math.floor(Math.random() * outcomes.length)];
+  const outcomes = userPosition.position.conditions[0].outcomes;
+  const outcomeString = outcomes ? outcomes[outcomeIndex - 1] : "";
+
+  const hasFinished = !!answer;
```
berteotti, line 43: We can use this instead of answer in line 47 and 116

app/components/CardBet.tsx
```diff
@@ -61,24 +96,28 @@ export const CardBet = ({ market }: CardBetProps) => {
...
                 alt="token logo"
-                size="xs"
+                className="size-3"
               />
             </div>
           </div>
-          <Button size="sm" colorScheme="success" variant="pastel">
-            Reedem
-          </Button>
+          {!!answer && (
```
berteotti, line 116: Let's change to `isWinner`

app/components/CardBet.tsx
Diogomartf: I think in that case is more clear what answer == outcomeIndex then hasFinished == outcomeIndex

models/Market.ts
```diff
@@ -0,0 +1,34 @@
...
+    this.isClosed = !!this.answer;
+  }
+
+  isWinner(outcomeIndex: number) {
+    return this.answer == outcomeIndex;
+  }
+
+  isLoser(outcomeIndex: number) {
+    return this.answer && this.answer != outcomeIndex;
+  }
```
berteotti, line 25: This can be derived from `isWinner`, so doesn't make much sense to have both. And they have different logic

models/Market.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  outcomeAmountString(outcomeIndex: number): string {
+    return this.answer
```
berteotti, line 28: Why not use `isClosed` here?

models/Position.ts
```diff
@@ -0,0 +1,17 @@
+import { Position } from "@/queries/conditional-tokens/types";
+
+export class PositionModel {
+  conditionId: string;
+  outcomeIndex: number;
+  outcomes: string[] | undefined | null;
```
berteotti, line 6: I think we shouldn't have undefined here. If there aren't outcomes, is null, otherwise is a string array.

models/Position.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  constructor(position: Position) {
+    this.conditionId = position.conditionIdsStr;
+    this.outcomeIndex = position.indexSets[0];
+    this.outcomes = position.conditions?.[0]?.outcomes;
+  }
+
+  outcomeString(): string {
+    return this.outcomes ? this.outcomes[this.outcomeIndex - 1] : "";
+  }
```
berteotti, line 16: This looks like a frontend thing. Doesn't look like a PositionModel problem. I would define `userOutcome` has a property that can be `this.outcomes[this.outcomeIndex - 1]` or `null` that we define in the constructor.

queries/conditional-tokens/queries.ts
```diff
@@ -0,0 +1,57 @@
+// @ts-nocheck
```
berteotti, line 1: Let's remove this

queries/conditional-tokens/queries.ts
```diff
@@ -0,0 +1,57 @@
...
+import { QueryUserPositionArgs } from "@/queries/conditional-tokens/types";
+import { gql, request } from "graphql-request";
+
+const CONDITIONAL_TOKENS_SUBGRAPH_URL =
+  "https://api.thegraph.com/subgraphs/name/gnosis/conditional-tokens-gc/";
```
berteotti, line 6: Let's remove this too

models/Market.ts
```diff
@@ -0,0 +1,34 @@
...
+  closingDate: Date;
+  answer: number;
+  isClosed: boolean;
+
+  constructor(market: FixedProductMarketMaker) {
+    this.data = market;
+    this.closingDate = new Date(+market.openingTimestamp * 1000);
+    this.answer =
+      market?.question?.currentAnswer &&
+      fromHex(market.question.currentAnswer, "number");
```
berteotti, line 15: Maybe this should be a condition and if there is no `currentAnswer`, answer should be null.

models/Market.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  isWinner(outcomeIndex: number) {
+    return this.answer == outcomeIndex;
+  }
+
+  isLoser(outcomeIndex: number) {
+    return this.answer && this.answer != outcomeIndex;
+  }
+
+  outcomeAmountString(outcomeIndex: number): string {
```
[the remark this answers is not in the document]
Diogomartf: true

models/Position.ts
Diogomartf: but then the type will shout cause, if positions has no conditions the "?" will return undefined.

models/Position.ts
Diogomartf: return null when no value, and handle that on componente.

models/Market.ts
Diogomartf: Not really as we have to check if the market is closed. 

For the winner we only have ot check if answer is equal to outcome index, as if answer is null (marked closed) it will be diff, so no need for that check.
I think this logic is needed, as otherwise will have to do this on the component level.

Diogomartf: I'll replace this.answer with this.isClosed

models/Market.ts
Diogomartf: 👍

models/Position.ts
Diogomartf: but we could counter that, but also position.conditions[0].outcomes can come as undefined. So we would have to handle that case, to make it null. Seems not to be worth it.

models/Market.ts
Diogomartf: true

models/Position.ts
berteotti: > but then the type will shout cause, if positions has no conditions the "?" will return undefined.

So we can remove the null in this case?

berteotti:
I have a bet that already closed but it's not showing correctly on the UI.

- It says "Pontential Win"
- Doesn't have a gradient background

image

berteotti (approved):

PR #18, opened 2024-04-25
My bets page - Empty states, Tabs logic

Diogomartf:
- add empty state
- add now wallet connected state
- add tabs logic
- update connect button to accept variant and color scheme props

image

No wallet connected state 
image


app/my-bets/page.tsx
```diff
@@ -23,6 +29,19 @@ export default function MyBetsPage() {
...
+  const userPositions = data?.userPositions ?? [];
+  const emptyData = data && data.userPositions.length == 0;
+
+  const filterActiveBets = (userPositions: UserPosition[]) =>
+    userPositions?.filter(
+      position => position.position.conditions[0].resolved === false
+    );
+
+  const filterClompleteBets = (userPositions: UserPosition[]) =>
+    userPositions?.filter(position => position.position.conditions[0].resolved);
```
berteotti, line 41: No need for a function here, we can filter directly with `userPositions`.
```suggestion
  const filterActiveBets = userPositions?.filter(
      position => position.position.conditions[0].resolved === false
    );

  const filterClompleteBets = userPositions?.filter(position => position.position.conditions[0].resolved);
```

app/my-bets/page.tsx
```diff
@@ -36,40 +55,72 @@ export default function MyBetsPage() {
...
                   ))
                 )}
               </TabPanel>
               <TabPanel>
-                <div className="bg-surface-primary-accent-1 p-5 rounded-4">
-                  in development..
-                </div>
+                {isLoading ? (
+                  <LoadingBets />
+                ) : filterClompleteBets.length === 0 ? (
```
berteotti, line 93: Here we are checking `filterClompleteBets` instead of `filterActiveBets`.

app/my-bets/page.tsx
```diff
@@ -36,40 +55,72 @@ export default function MyBetsPage() {
...
                 </div>
               </TabPanel>
-              <TabPanel>
-                <div className="bg-surface-warning-accent-1 p-5 rounded-4">
-                  Come back in a week!
-                </div>
+              <TabPanel className="space-y-4">
+                {isLoading ? (
+                  <LoadingBets />
+                ) : filterClompleteBets.length === 0 ? (
```
berteotti, line 113: If `filterClompleteBets` becomes an array this is correct. Otherwise this must change.

app/my-bets/page.tsx
Diogomartf: true, fixed.

models/Market.ts
```diff
@@ -4,15 +4,15 @@ import { fromHex } from "viem";
...
 
   constructor(market: FixedProductMarketMaker) {
     this.data = market;
     this.closingDate = new Date(+market.openingTimestamp * 1000);
-    this.answer = market.question?.currentAnswer
-      ? fromHex(market.question.currentAnswer, "number")
-      : null;
+    this.answer =
+      market?.question?.currentAnswer &&
+      fromHex(market.question.currentAnswer, "number");
```
berteotti, line 15: What happens if there isn't currentAnswer?

app/my-bets/page.tsx
```diff
@@ -23,6 +29,19 @@ export default function MyBetsPage() {
...
+  const userPositions = data?.userPositions ?? [];
+  const emptyData = data && data.userPositions.length == 0;
+
+  const filterActiveBets = userPositions?.filter(
+    position => position.position.conditions[0].resolved === false
+  );
+
+  const filterClompleteBets = userPositions?.filter(
+    position => position.position.conditions[0].resolved
+  );
```
berteotti, line 41: Doesn't userPositions always has a value? No need to conditional check if it is always set with a value.

berteotti (approved):
Approved but left a comment for a small improvement

PR #19, opened 2024-04-26
feat: improve swapbox flow

berteotti:
PR in draft.

Missing: 

 - fix modal on swapr-ui
 - transaction success modal is from swapr-ui and not presagio

image


Diogomartf:
Looking good.

One thing I noticed was when the tx is successfuly closing the "success" modal and the "confirm" modal is still open, which is odd. When tx is submitted we should close that modal.

app/components/ConfirmTrade.tsx
```diff
@@ -0,0 +1,327 @@
...
+      });
+      onApprove();
+    } catch (error) {
+      console.error(error);
+    } finally {
+      setIsTxLoading(false);
+    }
+  };
+
+  const submitBet = async () => {
```
Diogomartf, line 125:
```suggestion
  const submitBuyBet = async () => {
```

this rename indicates the method "buy" it makes it consisten with `submitSellBet`

Diogomartf:
when I'm disconnected and try to connect on a market through swapbox, It goes to error view.

image

constants/index.ts
```diff
@@ -1,8 +1,18 @@
+import { Token } from "@/entities";
+
 export const ChainId = {
   ETHEREUM: 1,
   GNOSIS: 100,
 };
 
+export const WXADI = new Token(
```
Diogomartf, line 8: TYPO

```suggestion
export const WXDAI = new Token(
```

PR #21, opened 2024-04-30
Add theme switcher with Popover and improve Navbar

Diogomartf:
image
image


app/components/ui/Navbar.tsx
```diff
@@ -16,31 +33,58 @@ export const Navbar = () => {
...
       </div>
     </nav>
   );
 };
+
+const SettingsPopover = ({
+  setTheme,
+}: {
+  setTheme: (theme: string) => void;
+}) => (
```
berteotti, line 62: No need to pass props to use `setTheme`, we can call the hook inside this component. And remove setTheme from the parent component where isn't being used.

app/components/ui/Navbar.tsx
```diff
@@ -16,31 +33,58 @@ export const Navbar = () => {
...
+    <PopoverContent className="max-w-md px-0">
+      <div className="space-y-2">
+        <div className="flex items-center mx-4 text-text-low-em">
+          <p className="font-bold text-xs">Theme settings</p>
+        </div>
+        <TabGroup
+          as="div"
+          onChange={(index: number) =>
+            console.log("Changed selected tab to:", index)
+          }
```
berteotti, line 76: Can we call `setTheme` here? If we can't option a value here, I think this isn't the right component. A closer component to RadioGroup makes more sense here.

berteotti:
The UI is different from the figma design. Not sure if we want to go forward with this first, but I like the figma design.

image

app/components/ui/Navbar.tsx
berteotti: Looking closely, at both design and logical level this isn't the right component for this. I think we should create a RadioGroup and use it here.

There are two options:
1. Commit this and work on a RadioGroup later
2. Work on a RadioGroup component, update this branch and then commit

Let me know what you think

app/components/ConnectButton.tsx
```diff
@@ -43,14 +43,16 @@ const CustomConnectButton = ({
     const balanceValue = formatEther(balanceData.value);
 
     // Regex to show only 3 digits after the decimal point
```
berteotti, line 45: Update the comment too 😄

Diogomartf:
> The UI is different from the figma design. Not sure if we want to go forward with this first, but I like the figma design.
> 
> image

Yeah it's the `ToggleGroup`

app/components/ui/Navbar.tsx
Diogomartf: You are right, we have `ToggleGroup` that does this (we choose to name it toggle group instead of radio).

berteotti (approved):

PR #22, opened 2024-04-30
feat: add popover to selecet bet outcome

berteotti:
- Add outcome picker, a popover to select bet outcomes.

image


Diogomartf (approved):
Looks great! Good job.

PR #24, opened 2024-05-02
fix: improve market outcome readability

berteotti:
- Improves Market outcome redability
- Fixes SettingsPopover hydratation problem

image


ElRodrigote (approved):

Diogomartf:
Related to truncated ui, we can't show truncated next to slash.

"...-"

image

It looks like is negative number.
Maybe inverting the order can fix this, but has to be tried out.

berteotti:
> Related to truncated ui, we can't show truncated next to slash.
> 
> "...-"
> 
> image
> 
> It looks like is negative number. Maybe inverting the order can fix this, but has to be tried out.

Removed the hyphen to make it clear it isn't negative. I think the most important info is the outcome name, percentage being a secondary info.

PR #26, opened 2024-05-08
feat: finished bet state on market detail

berteotti:
image

### DIFF
--- a/app/components/Swapbox.tsx
+++ b/app/components/Swapbox.tsx
@@ -196,7 +196,7 @@ export const Swapbox = ({ market }: { market: FixedProductMarketMaker }) => {
 
   return (
     <>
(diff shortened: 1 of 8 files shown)


ElRodrigote (commented):
So far looking good!

app/components/Swapbox.tsx
```diff
@@ -22,6 +22,7 @@ import {
 import { ConfirmTrade } from "./ConfirmTrade";
 import { ModalId, useModalContext } from "@/context/ModalContext";
 import { WXDAI } from "@/constants";
+import { MarketModel } from "../../models";
```
Diogomartf, line 25: can we use "@/models" so it will works without depending on the path

PR #29, opened 2024-05-13
Improve folder structure and name

Diogomartf:
- rename model -> contracts (as they are methods to interact with contracts)
- move models into entities and merge them
- rename PositionModel and MarketModel to Position and Market.


berteotti (commented):
Looking good but build is failing

README.md
```diff
@@ -1,4 +1,5 @@
 # Next dapp starter kit
+
 This is a template meant to kickstart dapps.
 
 ![Next dapp starter kit](/public/next-dapp-starter-kit.png)
```
ElRodrigote, line 5: Should we remove this?

app/components/ConfirmTrade.tsx
```diff
@@ -13,12 +13,14 @@ import { SLIPPAGE, SwapDirection, SwapState } from ".";
 import { useState } from "react";
 import { waitForTransactionReceipt, writeContract } from "wagmi/actions";
 import MarketABI from "@/abi/market.json";
-import { CONDITIONAL_TOKEN_CONTRACT_ADDRESS } from "@/model/conditionalTokens";
+import {
+  CONDITIONAL_TOKEN_CONTRACT_ADDRESS,
+  useReadCalcSellAmount,
```
ElRodrigote, line 18: Minor, non-blocking
I'd move all these variants of WAGMI's `useReadContract` into another folder, maybe something called `hooks` or something like that, as they're not contracts nor markets; they're utility functions.

WDYT?

ElRodrigote (approved):

app/components/ConfirmTrade.tsx
Diogomartf: They are related to contracts, that's why the name, but yes they are also only hooks. So hooks/contracts seems to do the trick. Thanks!

README.md
Diogomartf: No, I think it's okay, we should that we used our dapp starter kit

entities/markets/trade.ts
berteotti, line 1: Not sure if this should live under entities, but can be here for now

berteotti (approved):
Looking good

entities/markets/trade.ts
Diogomartf: Yes, not sure either, but let it be for a while and will see.

PR #30, opened 2024-05-14
feat: improve market user bets

berteotti:
- Shows user bets on market detail
- Allows bet claiming inside market detail

image


Diogomartf (approved):
this looks great, good job.

PR #31, opened 2024-05-16
feat: add search and filtering on markets page

berteotti:
image

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -8,37 +8,128 @@ import {
   getMarkets,
 } from "@/queries/omen";
 import { CardMarket, LoadingCardMarket } from "@/app/components";
+import {
+  Button,
+  Icon,
+  Input,
+  Popover,
+  PopoverContent,
+  PopoverTrigger,
+} from "swapr-ui";
+import { useState } from "react";
+import { useShowClientUI, useDebounce } from "@/hooks";
+import { cx } from "class-variance-authority";
+
+type FilterOption = {
+  name: string;
+  orderBy: FixedProductMarketMaker_OrderBy;
+};
+
+const filterOptions: FilterOption[] = [
+  {
+    name: "Newest",
+    orderBy: FixedProductMarketMaker_OrderBy.CreationTimestamp,
+  },
+  {
+    name: "Highest volume",
+    orderBy: FixedProductMarketMaker_OrderBy.CollateralVolume,
+  },
+];
+
+const ITEMS_PER_PAGE = 12;
 
 export default function HomePage() {
-  const { data: markets, isLoading } = useQuery({
-    queryKey: ["getMarkets"],
+  const showClientUI = useShowClientUI();
+  const [selectedOption, setSelectedOption] = useState(filterOptions[0]);
+  const [search, setSearch] = useState("");
+  const [isPopoverOpen, setPopoverOpen] = useState(false);
+  const debouncedSearch = useDebounce(search, 600);
+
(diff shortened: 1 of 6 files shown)


app/page.tsx
```diff
@@ -8,37 +8,126 @@ import {
...
       getMarkets({
-        first: 10,
+        first: ITEMS_PER_PAGE,
         skip: 0,
-        orderBy: FixedProductMarketMaker_OrderBy.CreationTimestamp,
+        orderBy: selectedOption.orderBy,
         orderDirection: OrderDirection.Desc,
+        title_contains_nocase: search,
       }),
   });
```
ElRodrigote, line 57: WDYT about adding a debounce here (for 500ms or 700ms) to reduce the amount of calls we do to query data?

app/page.tsx
```diff
@@ -8,37 +8,126 @@ import {
...
+          )}
         </div>
       </div>
+      {isLoading ? (
+        <div className="grid w-full grid-cols-1 gap-4 sm:grid-cols-2 xl:grid-cols-3 2xl:grid-cols-4">
+          {Array.from(Array(ITEMS_PER_PAGE).keys()).map((index) => (
+            <LoadingCardMarket key={index} />
+          ))}
+        </div>
+      ) : markets && markets.length > 0 ? (
```
ElRodrigote, line 118: Minor non blocking, this should be the same

```suggestion
      ) : markets?.length ? (
```

app/page.tsx
```diff
@@ -8,37 +8,126 @@ import {
...
+                    <div
+                      key={option.name}
+                      onClick={() => setSelectedOption(option)}
+                      className="flex items-center justify-start px-3 py-2 space-x-2 font-semibold cursor-pointer"
+                    >
+                      <Icon
+                        className={
+                          selectedOption.name !== option.name
+                            ? "invisible"
+                            : undefined
```
ElRodrigote, line 94: Minor, non-blocking: 
This should behave the same
```suggestion
                        className={cx({
                          invisible: selectedOption.name !== option.name,
                        })}
```

app/page.tsx
```diff
@@ -8,37 +8,126 @@ import {
...
+            <Button variant="pastel" className="space-x-2">
+              <p>{selectedOption.name}</p>
+              <Icon name="chevron-down" />
+            </Button>
+          )}
         </div>
       </div>
+      {isLoading ? (
+        <div className="grid w-full grid-cols-1 gap-4 sm:grid-cols-2 xl:grid-cols-3 2xl:grid-cols-4">
+          {Array.from(Array(ITEMS_PER_PAGE).keys()).map((index) => (
```
ElRodrigote, line 114: Minor, non-blocking:
This should behave the same, I think any of these options feel simpler and easier to read

```{[...Array(ITEMS_PER_PAGE)].map((index) => (```
```{Array(ITEMS_PER_PAGE).fill('').map((index) => (```
```{Array.from({ length: ITEMS_PER_PAGE}).map((index) => (```

ElRodrigote (approved):

Diogomartf:
Looking good,

image

I would make this 2 column on mobile, I think it will work better.

berteotti:
> I would make this 2 column on mobile, I think it will work better.

Made a different approach, preventing text from wrapping. I think it looks good this way, let me know what you think

Diogomartf (approved):

ElRodrigote (approved):

PR #33, opened 2024-05-22
Add unredeemed bets tab functionality

Diogomartf:
## Adds unredeemed tab
image

Description
- add unredeemed tab
- Also fixes bug not showing correct color when `isLoser`

### DIFF
--- a/app/components/CardBet.tsx
+++ b/app/components/CardBet.tsx
@@ -4,7 +4,6 @@ import { useQuery } from "@tanstack/react-query";
 import { cx } from "class-variance-authority";
 import Link from "next/link";
 import { useAccount, useConfig } from "wagmi";
-import { formatEther } from "viem";
 
 import { Button, Logo, Tag } from "swapr-ui";
 import { Card } from "@/app/components/ui";
@@ -12,7 +11,12 @@ import { Card } from "@/app/components/ui";
 import { UserPosition } from "@/queries/conditional-tokens/types";
 import { getConditionMarket, getMarketUserTrades } from "@/queries/omen";
 import { remainingTime } from "@/utils/dates";
-import { Market, Position, valueByTrade } from "@/entities";
+import {
+  Market,
+  Position,
+  tradesCollateralAmountUSDSpent,
+  tradesOutcomeBalance,
+} from "@/entities";
 import { redeemPositions } from "@/hooks/contracts";
 import { WXDAI } from "@/constants";
 import { waitForTransactionReceipt } from "wagmi/actions";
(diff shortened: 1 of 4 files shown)


app/my-bets/page.tsx
```diff
@@ -29,12 +37,85 @@ export default function MyBetsPage() {
...
+        } catch (error) {
+          console.log(error);
+        }
+      })
+    );
+
+    const filteredResults: UserPosition[] = results.filter(
+      (result): result is UserPosition => result !== undefined
+    );
+    console.log("filteredResults:", filteredResults);
```
ElRodrigote, line 105: Should we keep this?

app/my-bets/page.tsx
```diff
@@ -116,7 +198,7 @@ const BetsListPanel = ({
     <TabPanel className="space-y-4">
       {isLoading && <LoadingBets />}
       {!isLoading &&
-        bets.length &&
+        bets.length > 0 &&
```
ElRodrigote, line 201: Do we need this change? Feels redundant because `bets.length` should be truthy for any non-zero value and falsy for zero value.

entities/markets/market.ts
```diff
@@ -40,7 +40,7 @@ export class Market {
   }
 
   isLoser(index: number) {
-    return this.answer && this.answer !== index;
+    return this.answer !== null && this.answer !== index;
```
ElRodrigote, line 43: Maybe this also works? Or is there a reason to check if the value is explicitly null?
```suggestion
    return this?.answer !== index;
```

app/my-bets/page.tsx
Diogomartf: nope

app/my-bets/page.tsx
Diogomartf: we need this to boolean otherwise it prints the value

entities/markets/market.ts
Diogomartf: we need to check for null values, othis the same as the above and was not working, cause if it's null it will make this statement true.

app/my-bets/page.tsx
ElRodrigote: Ah I see, gotcha

ElRodrigote (approved):

ElRodrigote (approved):

app/components/CardBet.tsx
```diff
@@ -58,24 +61,16 @@ export const CardBet = ({ userPosition }: BetProps) => {
           outcomeIndex_in: [outcomeIndex],
         });
     },
-    enabled: !!address,
+    enabled: !!market?.data?.id,
```
berteotti, line 64: This should be enabled by both address and market being available.

app/my-bets/page.tsx
```diff
@@ -29,12 +36,81 @@ export default function MyBetsPage() {
   const filterActiveBets = userPositions.filter(
     position => position.position.conditions[0].resolved === false
   );
-  const filterClompleteBets = userPositions.filter(
+  const filterCompleteBets = userPositions.filter(
     position => position.position.conditions[0].resolved
   );
 
+  const fetchUnredeemedBets = useCallback(async () => {
```
berteotti, line 43: Probably we want to use `useMemo` here and not `useCallback`

entities/markets/trade.ts
```diff
@@ -1,8 +1,35 @@
...
     previousValue + newValue,
   [TradeType.Sell]: (previousValue: number, newValue: number) =>
     previousValue - newValue,
 };
+
+interface tradesProps {
+  fpmmTrades: FpmmTrade[] | undefined;
+}
+
+export const tradesOutcomeBalance = ({ fpmmTrades }: tradesProps) => {
```
berteotti, line 15: No need to receive an object if we have only 1 param.

entities/markets/trade.ts
```diff
@@ -1,8 +1,35 @@
...
+      const type = trade.type;
+      const collateralAmountUSD = parseFloat(
+        formatEther(trade.outcomeTokensTraded as bigint)
+      );
+      return valueByTrade[type](acc, collateralAmountUSD);
+    }, 0) ?? 0
+  );
+};
+
+export const tradesCollateralAmountUSDSpent = ({ fpmmTrades }: tradesProps) => {
```
berteotti, line 27: No need to receive an object if we have only 1 param.

PR #35, opened 2024-05-28
add filters & search to searchParams

Diogomartf:
Add filters to searchParams, this keeps the filters state on the url, it allows keeping the state on refresh and is sharable.

image

Add search to search params
image

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -45,7 +45,14 @@ export default function HomePage() {
   const router = useRouter();
   const showClientUI = useShowClientUI();
   const [isPopoverOpen, setPopoverOpen] = useState(false);
-  const searchParams = new URLSearchParams(window.location.search);
+
+  const searchParams =
+    typeof window !== "undefined"
+      ? new URLSearchParams(window.location.search)
+      : new URLSearchParams("/");
+
+  const [search, setSearch] = useState(searchParams.get("s") || "");
+  const debouncedSearch = useDebounce(search, SEARCH_DEBOUNCE_DELAY);
 
   const initialFilter = () => {
     const filterValueFromSearchParams = searchParams.get("f");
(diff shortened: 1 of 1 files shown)


app/page.tsx
```diff
@@ -57,6 +60,38 @@ export default function HomePage() {
       }),
   });
 
+  const handleSearch = (query: string) => {
+    const currentQuery = new URLSearchParams(window.location.search);
+    query ? currentQuery.set("s", query) : currentQuery.delete("s");
+    setSearch(query);
+    router.push(`${window.location.pathname}?${currentQuery.toString()}`);
```
berteotti, line 67: Can we re write browser history instead of push? Otherwise we will have a big history after doing a search

app/page.tsx
```diff
@@ -57,6 +60,38 @@ export default function HomePage() {
...
+  };
+
+  const selectFilter = (option: FilterOption) => {
+    setSelectedOption(option);
+    setPopoverOpen(false);
+
+    const currentQuery = new URLSearchParams(window.location.search);
+    currentQuery.set("f", option.orderBy);
+    router.push(`${window.location.pathname}?${currentQuery.toString()}`);
```
berteotti, line 76: Can we re write browser history instead of push here too?

app/page.tsx
```diff
@@ -57,6 +60,38 @@ export default function HomePage() {
...
+    if (searchValueFromSearchParams) setSearch(searchValueFromSearchParams);
+
+    if (filterValueFromSearchParams) {
+      const matchedOption = filterOptions.find(
+        option => option.orderBy === filterValueFromSearchParams
+      );
+
+      if (matchedOption) setSelectedOption(matchedOption);
+    }
+  }, []);
```
berteotti, line 93: If we can remove this logic and have search params initialising `search` and `selectedOption` that would be better.

berteotti (approved):

PR #36, opened 2024-05-29
Improve NoBetsPage design and update names

Diogomartf:
**Screenshots**

image

image


app/components/ui/images/NoBetsImage.tsx
```diff
@@ -0,0 +1,223 @@
+import { useTheme } from "next-themes";
```
ElRodrigote, line 1: I'm not sure about mixing components AND images inside a `components` folder, I'd move the images created here into the `public/assets/images` that we already have 

For these 2 new images we don't need a `.tsx` file as we won't use any conditional style like the one in `ConnectWalletImage`

Diogomartf: we do use the image conditionally, so why have it on the assets/images when can be near where it's used?
I think it's better to be around, unless it's suppose to be used on different places but I can't see that happening on this case.

Diogomartf: Although It doesn't hurt to have it on images. I guess I'll change it.

Diogomartf: ok, it went into errors, rendering the image 🤷‍♂️  So I guess I'll leave it as it is.

berteotti (approved):

PR #39, opened 2024-06-04
Add TxContext to abstract writeContract with TX modal - submitTx

Diogomartf:
### TxProvider

this provider wraps wagmi's `writeContract` method with the `TransactionModal` to show feedback from the written transaction: loading/success/error and also links to explorer.

**API**
`submitTx` - call writeContract, shows modal with tx

**TransactionModal**
Modal with transaction feedback 
- tx state: loading/success/error
- tx explorer link

image

image

### DIFF
--- a/app/components/ConfirmTrade.tsx
+++ b/app/components/ConfirmTrade.tsx
@@ -1,4 +1,4 @@
-import { ModalId, useModalContext } from "@/context/ModalContext";
+import { ModalId, useModal } from "@/context/ModalContext";
 import {
   Button,
   Dialog,
@@ -44,7 +44,7 @@ export const ConfirmTrade = ({
   tokenAmountIn,
   tokenAmountOut,
 }: ConfirmTradeProps) => {
-  const { isModalOpen, closeModal } = useModalContext();
+  const { isModalOpen, closeModal } = useModal();
   const { submitTx } = useTx();
 
   const { data: sellAmount } = useReadCalcSellAmount(
--- a/app/components/ConfirmTrade.tsx
+++ b/app/components/ConfirmTrade.tsx
@@ -10,8 +10,6 @@ import {
   Icon,
 } from "swapr-ui";
 import { SLIPPAGE, SwapDirection, SwapState } from ".";
-import { useState } from "react";
-import { waitForTransactionReceipt, writeContract } from "wagmi/actions";
 import MarketABI from "@/abi/market.json";
 import {
   CONDITIONAL_TOKEN_CONTRACT_ADDRESS,
@@ -19,10 +17,9 @@ import {
 } from "@/hooks/contracts";
 import ConditionalTokensABI from "@/abi/conditionalTokens.json";
 import { addFraction, removeFraction } from "@/utils/price";
-import { useConfig } from "wagmi";
-import { Address, erc20Abi, formatEther, parseEther } from "viem";
+import { Abi, Address, erc20Abi, formatEther, parseEther } from "viem";
 import { WXDAI } from "@/constants";
-import { TransactionModal } from "./TransactionModal";
+import { useTx } from "@/context";
 
 const ROUNDING_PRECISON = 0.00000000001;
 
@@ -47,10 +44,9 @@ export const ConfirmTrade = ({
   tokenAmountIn,
   tokenAmountOut,
 }: ConfirmTradeProps) => {
-  const config = useConfig();
-  const { isModalOpen, openModal, closeModal } = useModalContext();
-  const [txHash, setTxHash] = useState("");
-  const [isTxLoading, setIsTxLoading] = useState(false);
+  const { isModalOpen, closeModal } = useModalContext();
+  const { submitTx } = useTx();
+
   const { data: sellAmount } = useReadCalcSellAmount(
     marketId,
     formatEther(tokenAmountOut || BigInt(0)),
@@ -59,10 +55,6 @@ export const ConfirmTrade = ({
 
   const closeBetModal = () => {
     closeModal(ModalId.CONFIRM_SWAP);
-  };
-
-  const openTxModal = () => {
-    openModal(ModalId.WAITING_TRANSACTION);
   };
 
   const amountWei = parseEther(tokenAmountIn);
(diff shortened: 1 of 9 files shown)


app/components/TransactionModal.tsx
```diff
@@ -10,6 +10,7 @@ import {
 } from "swapr-ui";
 import { ModalId, useModalContext } from "@/context/ModalContext";
 import Image from "next/image";
+import { useEffect } from "react";
```
Diogomartf, line 13: remove useEffect

context/Web3Context.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { openModal } = useModalContext();
+
+  const writeContractShowModal = async (
+    config: Config,
+    args: WriteContractParameters
+  ): Promise<string | undefined> => {
+    setIsTxLoading(true);
+
+    try {
+      const txHash = await wagmiWriteContract(config, args);
```
berteotti, line 36: We call `config` inside this Provider. Make sure to call the right config file (`providers/config`, maybe we should change where we declare this file) or using `useConfig`

context/Web3Context.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const [txHash, setTxHash] = useState<string>("");
+  const [isTxLoading, setIsTxLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+
+  const { openModal } = useModalContext();
+
+  const writeContractShowModal = async (
+    config: Config,
+    args: WriteContractParameters
+  ): Promise<string | undefined> => {
```
berteotti, line 32: Let's not return undefined, let's throw on error.

context/Web3Context.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+      await waitForTransactionReceipt(config, {
+        hash: txHash,
+      });
+
+      return txHash;
+    } catch (error) {
+      setIsError(true);
+      console.error(error);
+      return undefined;
```
berteotti, line 49: It's better to throw an error here.

context/Web3Context.tsx
berteotti: I think just returning a string works fine.

context/Web3Context.tsx
berteotti: Or maybe do nothing has we are dealing with the error already.

context/Web3Context.tsx
```diff
@@ -0,0 +1,71 @@
...
+    config: Config,
+    args: WriteContractParameters
+  ) => Promise<string | undefined>;
+}
+
+export const Web3Context = createContext<Web3ContextProps>({
+  writeContractShowModal: async (config, args) => "",
+});
+
+export const Web3Provider = ({ children }: PropsWithChildren) => {
```
berteotti, line 22: This Provider name is very generic. Let's call it `TransactionProvider`.

context/Web3Context.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const [txHash, setTxHash] = useState<string>("");
+  const [isTxLoading, setIsTxLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+
+  const { openModal } = useModalContext();
+
+  const writeContractShowModal = async (
```
berteotti, line 29: I don't think this is a good name. I would call it, `submitTransaction`, only `writeContract`, or `startTransactionFlow`. Something around that.

context/Web3Context.tsx
berteotti: I think we need to throw an error, so we have to catch it. By catching we prevent calling `onApprove` if something goes wrong, for example. https://github.com/SwaprHQ/presagio/pull/39/files#diff-c24360049586ecf1a45d448ed999a5f309471ae6ca58880e97df385e7a6da1abR87

app/components/ConfirmTrade.tsx
```diff
@@ -48,9 +46,9 @@ export const ConfirmTrade = ({
   tokenAmountOut,
 }: ConfirmTradeProps) => {
   const config = useConfig();
-  const { isModalOpen, openModal, closeModal } = useModalContext();
-  const [txHash, setTxHash] = useState("");
-  const [isTxLoading, setIsTxLoading] = useState(false);
+  const { isModalOpen, closeModal } = useModalContext();
+  const { writeContractShowModal } = useWeb3Context();
```
berteotti, line 50: Here we should call is something like `useTransactionContext`or just `useTransaction`.

context/Web3Context.tsx
Diogomartf: Ok let's go with `TxProvider`

context/Web3Context.tsx
Diogomartf: Let's go with `submitTx`

context/Web3Context.tsx
Diogomartf: Yes, true, updated.

app/components/TransactionModal.tsx
```diff
@@ -54,6 +56,18 @@ export const TransactionModal = ({
...
+                <div className="flex flex-col items-center space-y-2">
+                  <p className="text-2xl font-semibold text-text-high-em">
+                    There was an error.
+                  </p>
+                  <p className="font-semibold text-center text-text-low-em text-md max-w-80">
+                    Unfortunately the transaction was not completed.
+                  </p>
+                </div>
+              </>
             ) : (
```
ElRodrigote, line 71: WDYT about simplifying the readability a bit by using short-circuit instead of multi-ternary operators? Something like:
```ts
{isError && ()}
{!isError && ()}
```

app/components/TransactionModal.tsx
```diff
@@ -71,19 +85,26 @@ export const TransactionModal = ({
...
-            size="lg"
-          >
-            <>
-              View in explorer <Icon name="arrow-top-right" />
-            </>
-          </Button>
+          {txHash && (
+            <a
+              href={`https://gnosisscan.io/tx/${txHash}`}
+              target="_blank"
```
ElRodrigote, line 91: Please prefer using `noopener` and `noreferrer` as extra security layers when using `target="_blank"`

https://medium.com/@halilatilla/target-blank-rel-noopener-noreferrer-b43e9ec05ade

Diogomartf: Yes, we should add the rel. Thanks

app/components/TransactionModal.tsx
Diogomartf: We can do that 👍

berteotti (approved):

PR #41, opened 2024-06-14
Allow filtering markets by categories (functionality)

Diogomartf:
This PR takes care of the initial functionality of categories. 
It allows filtering markets by a category at a time.

The design from Akash uses Tabs component but Tabs implies that we render content related to the tab, in this case we just want to choose an option from several options which is our `ToggleGroup` component does. 

So we'll go with that and then we can add a diff component to swapr-ui or just have diff design option to ToggleGroup, we should explore options.

**Current design**
image

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -16,6 +16,8 @@ import {
   Popover,
   PopoverContent,
   PopoverTrigger,
+  ToggleGroup,
+  ToogleGroupOption,
 } from "swapr-ui";
 import { useState } from "react";
 import { useShowClientUI, useDebounce } from "@/hooks";
@@ -43,10 +45,24 @@ const ITEMS_PER_PAGE = 12;
 const SEARCH_DEBOUNCE_DELAY = 600;
 const DEFAULT_FILTER_OPTION = filterOptions[0];
 
+enum Categories {
+  BUSINESS = "business",
+  CRYPTO = "crypto",
+  ECONOMY = "economy",
+  ENTERTAINMENT = "entertainment",
+  INTERNATIONAL = "international",
+  POLITICS = "politics",
+  SPORTS = "sports",
+  TECHNOLOGY = "technology",
+}
+
+type CategoryOptions = Categories | "";
+
 export default function HomePage() {
   const router = useRouter();
   const showClientUI = useShowClientUI();
   const [isPopoverOpen, setPopoverOpen] = useState(false);
+  const [category, setCategory] = useState<CategoryOptions>("");
 
   const searchParams =
     typeof window !== "undefined"
@@ -62,7 +78,7 @@ export default function HomePage() {
 
     return (
       filterOptions.find(
-        (option) => option.orderBy === filterValueFromSearchParams
+        option => option.orderBy === filterValueFromSearchParams
       ) || DEFAULT_FILTER_OPTION
     );
   };
@@ -79,7 +95,13 @@ export default function HomePage() {
   const [page, setPage] = useState(initialPage());
 
   const { data, isLoading } = useQuery({
-    queryKey: ["getMarkets", debouncedSearch, selectedOption.orderBy, page],
+    queryKey: [
+      "getMarkets",
+      debouncedSearch,
+      selectedOption.orderBy,
+      page,
+      category,
+    ],
     queryFn: async () =>
       getMarkets({
         first: ITEMS_PER_PAGE,
@@ -88,6 +110,7 @@ export default function HomePage() {
         orderDirection: OrderDirection.Desc,
         title_contains_nocase: debouncedSearch,
         creator_in: AI_AGENTS_ALLOWLIST,
+        category_contains: category,
       }),
   });
 
@@ -124,7 +147,13 @@ export default function HomePage() {
   const nextPage = page * ITEMS_PER_PAGE;
 
   const { data: marketsNextPage } = useQuery({
-    queryKey: ["getMarkets", debouncedSearch, selectedOption.orderBy, nextPage],
+    queryKey: [
+      "getMarkets",
+      debouncedSearch,
+      selectedOption.orderBy,
+      nextPage,
+      category,
+    ],
     queryFn: async () =>
       getMarkets({
         first: ITEMS_PER_PAGE,
@@ -133,6 +162,7 @@ export default function HomePage() {
         orderDirection: OrderDirection.Desc,
         title_contains_nocase: debouncedSearch,
         creator_in: AI_AGENTS_ALLOWLIST,
+        category_contains: category,
       }),
   });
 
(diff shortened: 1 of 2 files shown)


app/page.tsx
```diff
@@ -144,15 +183,35 @@ export default function HomePage() {
...
+        <h1 className="text-2xl font-semibold text-white capitalize">
+          🔮 {category ? category : "All"}
+        </h1>
+      </div>
       <div className="flex flex-col justify-between w-full gap-5 md:flex-row">
-        <h1 className="text-2xl font-semibold text-white">🔮 All markets</h1>
+        <ToggleGroup
+          value={category}
+          onChange={setCategory}
+          // className="overflow-x-scroll"
```
ElRodrigote, line 196: Should we keep this `className`?

app/page.tsx
```diff
@@ -144,15 +183,35 @@ export default function HomePage() {
   const showPaginationButtons = hasMoreMarkets || page !== 1;
 
   return (
-    <div className="justify-center px-6 mt-12 space-y-12 md:px-10 lg:px-20 xl:px-40 md:flex md:flex-col md:items-center">
+    <div className="justify-center px-6 mt-12 space-y-8 md:px-10 lg:px-20 xl:px-40 md:flex md:flex-col md:items-center">
+      <div className="w-full">
+        <h1 className="text-2xl font-semibold text-white capitalize">
+          🔮 {category ? category : "All"}
+        </h1>
+      </div>
```
ElRodrigote, line 191: Maybe we can remove that extra `<div>`

```suggestion
      <h1 className="text-2xl font-semibold text-white w-full capitalize">
        🔮 {category ? category : "All"}
      </h1>
```

app/page.tsx
```diff
@@ -43,10 +45,33 @@ const ITEMS_PER_PAGE = 12;
...
+const categories: CategoryOptions[] = [
+  "technology",
+  "crypto",
+  "politics",
+  "economy",
+  "business",
+  "international",
+  "sports",
+  "entertainment",
+];
```
ElRodrigote, line 68: Maybe we can use an `enum` here as both iterable object and type

```suggestion
enum CATEGORIES {
  BUSINESS = "business",
  CRYPTO = "crypto",
  ECONOMY = "economy",
  ENTERTAINMENT = "entertainment",
  INTERNATIONAL = "international",
  POLITICS = "politics",
  SPORTS = "sports",
  TECHNOLOGY = "technology",
}
```

app/page.tsx
```diff
@@ -43,10 +45,33 @@ const ITEMS_PER_PAGE = 12;
...
+
 export default function HomePage() {
   const router = useRouter();
   const showClientUI = useShowClientUI();
   const [isPopoverOpen, setPopoverOpen] = useState(false);
+  const [category, setCategory] = useState<CategoryOptions>("");
```
ElRodrigote, line 74: If you decide to use the enum, this change may be needed:

```suggestion
  const [category, setCategory] = useState< CATEGORIES>();
```

app/page.tsx
```diff
@@ -88,6 +119,7 @@ export default function HomePage() {
         orderDirection: OrderDirection.Desc,
         title_contains_nocase: debouncedSearch,
         creator_in: AI_AGENTS_ALLOWLIST,
+        category: category,
```
ElRodrigote, line 122: We could reduce verbosity here

```suggestion
        category,
```

app/page.tsx
```diff
@@ -133,6 +171,7 @@ export default function HomePage() {
...
```
ElRodrigote, line 174: And here

```suggestion
        category,
```

app/page.tsx
```diff
@@ -144,15 +183,35 @@ export default function HomePage() {
   const showPaginationButtons = hasMoreMarkets || page !== 1;
 
   return (
-    <div className="justify-center px-6 mt-12 space-y-12 md:px-10 lg:px-20 xl:px-40 md:flex md:flex-col md:items-center">
+    <div className="justify-center px-6 mt-12 space-y-8 md:px-10 lg:px-20 xl:px-40 md:flex md:flex-col md:items-center">
+      <div className="w-full">
+        <h1 className="text-2xl font-semibold text-white capitalize">
+          🔮 {category ? category : "All"}
```
ElRodrigote, line 189: With the previous change with the enum, we can simplify this to use a null coalescing operator

```suggestion
          🔮 {category ?? "All"}
```

app/page.tsx
```diff
@@ -144,15 +183,35 @@ export default function HomePage() {
...
+      </div>
       <div className="flex flex-col justify-between w-full gap-5 md:flex-row">
-        <h1 className="text-2xl font-semibold text-white">🔮 All markets</h1>
+        <ToggleGroup
+          value={category}
+          onChange={setCategory}
+          // className="overflow-x-scroll"
+        >
+          <ToogleGroupOption value="">All</ToogleGroupOption>
+          {categories.map(category => (
```
ElRodrigote, line 199: If you decide to use the enum, this change may be needed:

```suggestion
          {Objects.values(CATEGORIES).map(category => (
```

app/page.tsx
Diogomartf: YES, great! Thanks

app/page.tsx
Diogomartf: true

app/page.tsx
Diogomartf: Sweeeet, thanks!

app/page.tsx
Diogomartf: Yes, good catch.

app/page.tsx
Diogomartf: I need to update the component to receive the className, but currently it does not add classname but replaces all the classes with the one we pass.

app/page.tsx
ElRodrigote: Nono it will only work if you apply all the other changes (use the enum instead the type and default the useState to `useState()` instead of `useState("")`

berteotti: Honestly I would remove this, is just "stealing" space. I would make filters bigger and easier to read instead.

queries/omen/markets.ts
```diff
@@ -85,6 +85,7 @@ const getMarketsQuery = gql`
     $orderDirection: String
     $title_contains_nocase: String
     $creator_in: [String]
+    $category: String
```
berteotti, line 88: let's give it the same name: $category_contains

```suggestion
    $category_contains: String
```

app/page.tsx
Diogomartf: @berteotti currently is better with the title. But it will be removed on the next iteration (when we change to correct element) so not important to have it super clean 😁

Diogomartf: @ElRodrigote I need the empty string. But as I said above it will be removed on the next iteration. So no need to be super clean.

queries/omen/markets.ts
Diogomartf: yeah, makes sense.

berteotti (approved):
I think adding search params to categories will improve UX. We can do it in a diff PR but here looks good to and isn't a big change.

app/page.tsx
```diff
@@ -144,15 +174,35 @@ export default function HomePage() {
...
+          value={category}
+          onChange={setCategory}
+          // className="overflow-x-scroll"
+        >
+          <ToogleGroupOption value={""}>All</ToogleGroupOption>
+          {Object.values(Categories).map(category => (
+            <ToogleGroupOption
+              key={category}
+              value={category}
+              className="capitalize"
```
berteotti, line 194: Can we add bolder text here? It looks very thin and hard to read.

ElRodrigote (approved):

Diogomartf:
> I think adding search params to categories will improve UX. We can do it in a diff PR but here looks good to and isn't a big change.

Yeah, we have to add it, but I prefer to do it on another PR.

app/page.tsx
Diogomartf: yes. But currently not working, will update `ToogleGroupOption` component to allow it.

PR #43, opened 2024-06-18
Add category to SearchParams

Diogomartf:
image

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -46,14 +46,14 @@ const SEARCH_DEBOUNCE_DELAY = 600;
 const DEFAULT_FILTER_OPTION = filterOptions[0];
 
 enum Categories {
-  BUSINESS = "business",
-  CRYPTO = "crypto",
-  ECONOMY = "economy",
-  ENTERTAINMENT = "entertainment",
-  INTERNATIONAL = "international",
-  POLITICS = "politics",
-  SPORTS = "sports",
   TECHNOLOGY = "technology",
+  CRYPTO = "crypto",
+  BUSINESS = "business",
+  POLITICS = "politics",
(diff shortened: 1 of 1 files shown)


app/page.tsx
```diff
@@ -62,13 +62,14 @@ export default function HomePage() {
...
   const showClientUI = useShowClientUI();
   const [isPopoverOpen, setPopoverOpen] = useState(false);
-  const [category, setCategory] = useState<CategoryOptions>("");
 
   const searchParams =
     typeof window !== "undefined"
       ? new URLSearchParams(window.location.search)
       : new URLSearchParams("/");
 
+  const [category, setCategory] = useState(searchParams.get("c") || "");
```
ElRodrigote, line 71: Is there a reason for us to use `c` instead `category` as search param?

ElRodrigote (approved):

app/page.tsx
Diogomartf: we are using only one letter on search params.

No big reason, I just found it less verbose, and for URLs usually I like them short. 

The trade off is you lose a bit of readability but I don't expect people to create the url themselves, and for the super curious once you learn it, it's easier to write.

berteotti (approved):
I think scrollbar can be smaller. Otherwise looks good. Also I think we have too many categories, at least in the same bar.

Diogomartf:
> I think scrollbar can be smaller. Otherwise looks good. Also I think we have too many categories, at least in the same bar.


> I think scrollbar can be smaller. Otherwise looks good. Also I think we have too many categories, at least in the same bar.

We will tweak the categories in the future, this is also a "transition" component, till we do the proper design. But it gets the functionality going.

PR #44, opened 2024-06-19
feat: add filter market by state

berteotti:
- Adds filter by market state. Created new queries for each diff state query.
- Adds new order filter.

image

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -2,16 +2,7 @@
 
 import { useQuery } from "@tanstack/react-query";
 import Link from "next/link";
-import {
-  FixedProductMarketMaker_Filter,
-  FixedProductMarketMaker_OrderBy,
-  OrderDirection,
-  getMarkets,
-  getMarketsClosedQuery,
-  getMarketsDisputeQuery,
-  getMarketsOpenQuery,
-  getMarketsPendingQuery,
-} from "@/queries/omen";
+import { getMarkets } from "@/queries/omen";
 import { CardMarket, LoadingCardMarket } from "@/app/components";
 import {
   Button,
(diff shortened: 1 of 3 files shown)


app/page.tsx
```diff
@@ -41,9 +46,59 @@ const filterOptions: FilterOption[] = [
...
+  {
+    name: "In Dispute",
+    key: "dispute",
+
+    state: {
+      isPendingArbitration: true,
+    },
+    query: getMarketsDisputeQuery,
+  },
+];
```
Diogomartf, line 96: it would be great if we could move this to another file place, not sure where.

app/page.tsx
```diff
@@ -62,6 +117,7 @@ export default function HomePage() {
   const router = useRouter();
   const showClientUI = useShowClientUI();
   const [isPopoverOpen, setPopoverOpen] = useState(false);
+  const [isStatePopoverOpen, setStatePopoverOpen] = useState(false);
```
ElRodrigote, line 120:
```suggestion
  const [toggleStatusPopover, setToggleStatusPopover] = useState(false);
```

app/page.tsx
```diff
@@ -41,9 +46,59 @@ const filterOptions: FilterOption[] = [
...
+const nowTimestamp = Math.floor(Date.now() / 1000);
+
+type StateFilterOption = {
+  name: string;
+  key: string;
+  query: string;
+  state: FixedProductMarketMaker_Filter;
+};
+
+const stateFilterOptions: StateFilterOption[] = [
```
ElRodrigote, line 58: I'd shorten a bit the naming and change a few things:

* I'd drop "Options" from the variables, using a plural implies it has multiple entries or options, and if we want to specify it's an array structure, we could use "List" instead of "Options".
e.g.: `StateFilters`, `StateFilterList`, `selectedState`
* I'd change "State" to "Status", mostly because I'm used to seeing (in codebases) "state" tied to either DOM or storage state.
e.g.: `StatusFilters`, `StatusFilterList`, `selectedStatus`

^These suggestions could lead to an easier-to-read code being less verbose, but it's just a personal opinion, I have no strong opinions about this specific.

ElRodrigote (approved):

app/page.tsx
```diff
@@ -251,6 +342,44 @@ export default function HomePage() {
               <Icon name="chevron-down" />
             </Button>
           )}
+
+          {showClientUI ? (
+            <Popover
+              open={isStatePopoverOpen}
+              onOpenChange={setStatePopoverOpen}
+            >
+              <PopoverTrigger>
```
ElRodrigote, line 351: I'd place this `asChild` prop in all the `PopoverTrigger`components in this page, so when we publish the updates on Button components, the error we receive in the console will disappear without new changes
```suggestion
              <PopoverTrigger asChild>
```

app/page.tsx
```diff
@@ -251,6 +342,44 @@ export default function HomePage() {
...
+                      <Icon
+                        className={cx({
+                          invisible: selectedStateOption.key !== option.key,
+                        })}
+                        name="tick-fill"
+                      />
+                      <p>{option.name}</p>
+                    </div>
+                  );
+                })}
```
ElRodrigote, line 374: Minor, non-blocking:
We can use implicit return to drop verbose code

```suggestion
                {stateFilterOptions.map((option) => (
                    <div
                      key={option.key}
                      onClick={() => selectStateFilter(option)}
                      className="flex items-center justify-start px-3 py-2 space-x-2 font-semibold cursor-pointer"
                    >
                      <Icon
                        className={cx({
                          invisible: selectedStateOption.key !== option.key,
                        })}
                        name="tick-fill"
                      />
                      <p>{option.name}</p>
                    </div>
                  ))}
```

app/page.tsx
berteotti: I think in this case we can call it "is" since it's a boolean value. "toggle" sounds like an action.

app/page.tsx
berteotti: That's right, thanks!

app/page.tsx
berteotti: Changed the naming. Let me know what you think

app/page.tsx
```diff
@@ -159,22 +180,28 @@ export default function HomePage() {
...
+          title_contains_nocase: debouncedSearch,
+          creator_in: AI_AGENTS_ALLOWLIST,
+          category_contains: category,
+          ...selectedStateOption.state,
+        },
+        selectedStateOption.query
+      ),
   });
-
+  console.log(marketsNextPage);
```
Diogomartf, line 204: log
```suggestion
```

Diogomartf (approved):

PR #45, opened 2024-06-19
on MarketDetail add ToggleGroup to change between "Bet" and "history"

Diogomartf:
Add the initial tabs for switching "bet" and "history".
This will use the component ToggleGroup, which reflect the design. In the near future we should add this design to Tabs as well.  

image

image

### DIFF
--- a/app/markets/MarketDetails.tsx
+++ b/app/markets/MarketDetails.tsx
@@ -3,19 +3,31 @@
 import { OutcomeBar, Swapbox } from "@/app/components";
 import { useQuery } from "@tanstack/react-query";
 import { getMarket } from "@/queries/omen";
-import { Button, IconButton, Tag } from "swapr-ui";
+import {
+  Button,
+  IconButton,
+  Tag,
+  ToggleGroup,
+  ToggleGroupOption,
+} from "swapr-ui";
 import { remainingTime } from "@/utils/dates";
-import Link from "next/link";
 import { Address } from "viem";
 import { Market } from "@/entities";
 import { UserBets } from "../components/UserBets";
 import { useRouter } from "next/navigation";
+import { useState } from "react";
 
 interface MarketDetailsProps {
   id: Address;
 }
 
+enum Tabs {
+  BET = "bet",
+  HISTORY = "history",
+}
+
 export const MarketDetails = ({ id }: MarketDetailsProps) => {
+  const [tab, setTab] = useState<Tabs>(Tabs.BET);
   const { data, error, isLoading } = useQuery({
     queryKey: ["getMarket", id],
     queryFn: async () => getMarket({ id }),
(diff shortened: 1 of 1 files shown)


app/markets/MarketDetails.tsx
```diff
@@ -65,9 +73,44 @@ export const MarketDetails = ({ id }: MarketDetailsProps) => {
...
+              </div>
+              <div className="w-full">
+                <ToggleGroupOption
+                  size="md"
+                  value={"history"}
+                  className="font-semibold flex justify-center"
+                >
+                  History
+                </ToggleGroupOption>
+              </div>
```
ElRodrigote, line 99: Minor refactor, non-blocking (feel free to dismiss)
```suggestion
              {["bet", "history"].map((tab) => (
                <MarketTab tabName={tab} />
              ))}
```

The `MarketTab` component would look something like:
```tsx
const MarketTab = ({ tabName }: { tabName: string }) => (
  <div className="w-full">
    <ToggleGroupOption
      size="md"
      value={tabName}
      className="capitalize font-semibold flex justify-center"
    >
      {tabName}
    </ToggleGroupOption>
  </div>
);
```

app/markets/MarketDetails.tsx
```diff
@@ -65,9 +73,44 @@ export const MarketDetails = ({ id }: MarketDetailsProps) => {
...
+                <Swapbox market={market} />
+              ) : (
+                <div className="p-4 text-center">Market closed</div>
+              )}
+            </div>
+          )}
+          {tab === "history" && (
             <div className="p-2">
-              <Swapbox market={market} />
+              <div className="p-4 text-center">Market history</div>
```
ElRodrigote, line 113:
```suggestion
              <div className="p-4 text-center">No History yet</div>
```

app/markets/MarketDetails.tsx
```diff
@@ -65,9 +73,44 @@ export const MarketDetails = ({ id }: MarketDetailsProps) => {
               <OutcomeBar market={market} />
             </div>
           </div>
-          {!marketModel.isClosed && (
+          <div className="px-4">
```
berteotti, line 76: I think is missing a bit of bottom padding so tabs are a bit more far away from the swapbox.

app/markets/MarketDetails.tsx
```diff
@@ -3,19 +3,27 @@
...
 import { UserBets } from "../components/UserBets";
 import { useRouter } from "next/navigation";
+import { useState } from "react";
 
 interface MarketDetailsProps {
   id: Address;
 }
 
 export const MarketDetails = ({ id }: MarketDetailsProps) => {
+  const [tab, setTab] = useState("bet");
```
berteotti, line 26: Let's create an enum for bet and history options

berteotti (commented):
I suggest updating the skeleton too

app/markets/MarketDetails.tsx
Diogomartf: Don't think we need a component, but I'll take the array. 👍

app/markets/MarketDetails.tsx
Diogomartf: this is just a placeholder at this stage, as we we're building by steps.

app/markets/MarketDetails.tsx
Diogomartf: I'll experiment

PR #46, opened 2024-06-20
Add tab to switch between activity and chart

Diogomartf:
image

### DIFF
--- a/app/markets/HistorySection.tsx
+++ b/app/markets/HistorySection.tsx
@@ -1,0 +1,53 @@
+import {
+  Icon,
+  TabBody,
+  TabGroup,
+  TabHeader,
+  TabPanel,
+  TabStyled,
+} from "swapr-ui";
+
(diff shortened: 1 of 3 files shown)


app/markets/HistorySection.tsx
```diff
@@ -0,0 +1,53 @@
...
+  TabStyled,
+} from "swapr-ui";
+
+export const HistorySection = () => {
+  return (
+    <div className="p-2">
+      <TabGroup
+        onChange={(index: number) =>
+          console.log("Changed selected tab to:", index)
+        }
```
berteotti, line 16: This onChange shouldn't be required, right? If it's not, it can be deleted.

berteotti (approved):

app/markets/HistorySection.tsx
Diogomartf: yes, it's optional

PR #47, opened 2024-06-20
Show activity on Market - list of user trades  (UI)

Diogomartf:
This is only the UI for activity

Uses [Shadcn Table](https://ui.shadcn.com/docs/components/table) table component
Added 2 utils functions to help make things more readable:
- `formatDateTime`
- `shortenAddress`

few comments:
- data is mocked
- will not attempt to show avatar based on address at this time.
- Detect AI will come later
- pagination will come later

image

### DIFF
--- a/app/components/ui/Table.tsx
+++ b/app/components/ui/Table.tsx
@@ -1,0 +1,119 @@
+import { cx } from "class-variance-authority";
+import * as React from "react";
+
+const Table = React.forwardRef<
+  HTMLTableElement,
+  React.HTMLAttributes<HTMLTableElement>
+>(({ className, ...props }, ref) => (
+  <div className="relative w-full overflow-auto">
+    <table
+      ref={ref}
+      className={cx("w-full caption-bottom text-sm", className)}
+      {...props}
+    />
+  </div>
+));
+Table.displayName = "Table";
+
+const TableHeader = React.forwardRef<
+  HTMLTableSectionElement,
+  React.HTMLAttributes<HTMLTableSectionElement>
+>(({ className, ...props }, ref) => (
+  <thead ref={ref} className={cx("[&_tr]:border-b", className)} {...props} />
+));
+TableHeader.displayName = "TableHeader";
+
+const TableBody = React.forwardRef<
+  HTMLTableSectionElement,
+  React.HTMLAttributes<HTMLTableSectionElement>
+>(({ className, ...props }, ref) => (
+  <tbody
+    ref={ref}
+    className={cx("[&_tr:last-child]:border-0", className)}
+    {...props}
+  />
+));
+TableBody.displayName = "TableBody";
+
+const TableFooter = React.forwardRef<
+  HTMLTableSectionElement,
+  React.HTMLAttributes<HTMLTableSectionElement>
+>(({ className, ...props }, ref) => (
+  <tfoot
+    ref={ref}
+    className={cx(
+      "border-t bg-muted/50 font-medium [&>tr]:last:border-b-0",
+      className
+    )}
+    {...props}
+  />
+));
+TableFooter.displayName = "TableFooter";
+
+const TableRow = React.forwardRef<
+  HTMLTableRowElement,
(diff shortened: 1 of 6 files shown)


app/components/ui/Table.tsx
berteotti, line 1: We can then migrate this to Swapr-ui, right?

utils/token.ts
```diff
@@ -0,0 +1,2 @@
+export const shortenAddress = (address: string) =>
+  `${address.slice(0, 4)}...${address.slice(-3)}`;
```
berteotti, line 2: Let's make it 5.
Or 6 and -4.

```suggestion
  `${address.slice(0, 5)}...${address.slice(-3)}`;
```

app/markets/ActivityTable.tsx
```diff
@@ -0,0 +1,88 @@
...
+    <Table>
+      <TableHeader>
+        <TableRow>
+          <TableHead className="text-text-low-em">Users</TableHead>
+          <TableHead className="text-text-low-em">Action</TableHead>
+          <TableHead className="text-text-low-em">Shares</TableHead>
+          <TableHead className="text-text-low-em text-right">Date</TableHead>
+        </TableRow>
+      </TableHeader>
+      <TableBody className="font-semibold">
```
berteotti, line 62: Text can be bigger, right?

app/components/ui/Table.tsx
ElRodrigote: Yeah, we could migrate the generic component to the monorepo :D

Diogomartf: Maybe, but currently it's just a straight copy from `shadcn/ui`, so no big deal.

app/markets/ActivityTable.tsx
Diogomartf: it can, but I actually like this size, but I can increase it a bit to try out.

Diogomartf: Added text-base (which increases font in 2px)
image

utils/token.ts
Diogomartf: Addresses don't give much info, it's just a small set of numbers to be recognisable as an address and maybe in some cases you identify it, like some of your wallets you know they start in some digits.

Either way I increase it a bit on both sides.

app/markets/HistorySection.tsx
```diff
@@ -27,13 +24,8 @@ export const HistorySection = () => {
...
         <TabBody className="my-4">
           <TabPanel>
-            <div className="bg-surface-surface-2 h-44 rounded-8">
-              <div className="flex items-center justify-center h-full">
-                <div className="flex flex-col items-center space-y-2">
-                  <Icon name="filter" size={24} />
-                  <p>Activity coming soon</p>
-                </div>
-              </div>
+            <div className="rounded-8">
```
ElRodrigote, line 27: WDYT about adding a `className` to the `ActivityTable` component and remove this `div` element here?

ElRodrigote (approved):

berteotti (approved):
Looking good to me

app/markets/ActivityTable.tsx
berteotti: I was suggesting 6 for the first one because we tend to memorize the first 4 digits of a certain (mostly personal) address. Last ones can be 3 to 4. We don't need to change it now, I just think we can improve and the box has space for it.

app/markets/ActivityTable.tsx
```diff
@@ -0,0 +1,88 @@
...
+  TableBody,
+  TableCell,
+  TableHead,
+  TableHeader,
+  TableRow,
+} from "@/app/components/ui/Table";
+import { formatDateTime, shortenAddress } from "@/utils";
+import { Tag } from "swapr-ui";
+
+const activities = [
```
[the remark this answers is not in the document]
Diogomartf: yes

app/markets/HistorySection.tsx
Diogomartf: we can remove this div 👍

PR #48, opened 2024-06-21
[PGIO-37][IMPROV] - Add a Prettier config on repo

ElRodrigote:
## Fixes: [PGIO-37](https://linear.app/swaprhq/issue/PGIO-37/check-a-prettier-config-on-repo)

# Description
* Installs and adds a series of Prettier rules to format our codebase
* Adds an ignore file to place folders and files to be ignored from this formatting
* Creates a VSCode extension suggestions file
* Formats the whole codebase
* Adds a tailwind plugin for Prettier, sorting the whole codebase classnames


.vscode/extensions.json
berteotti, line 1: Not sure about this. Should we add .vscode to the repo?

package.json
```diff
@@ -1,46 +1,48 @@
...
+  "name": "next-dapp-starter-kit",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint",
+    "code:check": "npx prettier . --check",
+    "code:format": "npx prettier . --write",
```
berteotti, line 11: Let's not have this for now. We can add them if we need. This aren't currently being used right?

.prettierrc
```diff
@@ -1,9 +1,9 @@
 {
   "trailingComma": "es5",
   "tabWidth": 2,
-  "arrowParens": "always",
+  "arrowParens": "avoid",
   "singleQuote": true,
   "quoteProps": "as-needed",
   "bracketSpacing": true,
-  "printWidth": 120
+  "printWidth": 90
```
berteotti: Why 90?

.vscode/extensions.json
ElRodrigote: The idea of this file is to suggest VSCode extensions to devs using VSCode (like me :P), we can remove it.

What this does is: when you open the project in VSCode IDE, the IDE will detect the extensions array and suggest you to install them (as opt-in).

The suggested extension is the official one for Prettier

package.json
ElRodrigote: I used them to format the codebase safely, but right now there is no pipeline or hook (like pre-commit or pre-push hooks) using them.

I'll remove them :D

.prettierrc
ElRodrigote: 90 is a kinda safe number for 14" Laptop viewports and above. Tried 120 on my laptop but had to scroll right :c

.vscode/extensions.json
berteotti: It works for me

berteotti (approved):

Diogomartf:
I prefer the double quote "" but as I'm more use to do it, but if it just swaps auto when I save, I'm okay with it.

Diogomartf (approved):

PR #52, opened 2024-06-26
feat: enable all markets

berteotti:
- Add creator filters
- Update logic to use different search params on the query

image


app/page.tsx
```diff
@@ -80,6 +88,19 @@ export default function HomePage() {
 
   const [selectedStateOption, setSelectedStateOption] = useState(initialStateFilter());
 
+  const initialCratorFilter = () => {
```
Diogomartf, line 91:
```suggestion
  const initialCreatorFilter = () => {
```

Diogomartf (approved):

PR #53, opened 2024-06-26
Add activity table to market details

Diogomartf:
- add the graph query
- add table component
- add table loading
- gnosisExplorerAddress to utils token

**preview**
image

In following PRs we'll add:
- pagination
- small design improvements

### DIFF
--- a/app/markets/ActivityTable.tsx
+++ b/app/markets/ActivityTable.tsx
@@ -6,49 +6,43 @@ import {
   TableHeader,
   TableRow,
 } from '@/app/components/ui/Table';
-import { formatDateTime, shortenAddress } from '@/utils';
-import { Tag } from '@swapr/ui';
+import { TransactionType, getMarketTransactions } from '@/queries/omen';
+import {
+  formatDateTime,
+  formatEtherWithFixedDecimals,
+  getGnosisAddressExplorerLink,
+  shortenAddress,
+} from '@/utils';
+import { Tag, TagColorSchemeProp } from '@swapr/ui';
+import { useQuery } from '@tanstack/react-query';
 
-const activities = [
-  {
-    user: '0x8a05649a4fa186d032376f6d34f6bf40ca0168ae',
-    action: 'Yes',
-    shares: '1.24',
-    date: 1718999911,
-  },
-  {
-    user: '0x1b3bbecdcb76594e1dc5f7550263cd5ea4a6ab6d',
-    action: 'No',
-    shares: '2.3',
-    date: 1718999111,
-  },
-  {
-    user: '0x7532a3ce4d80cc3fb71d69a5329f76f9f375aa27',
-    action: 'Yes',
-    shares: '32.2',
-    date: 1718912111,
-  },
-  {
-    user: '0x2fe3e7422f35ecbb9dca542279fb2ee172faab32',
-    action: 'Yes',
-    shares: 23.1,
-    date: 1718903353,
-  },
-  {
-    user: '0x269e4c1cb7d8586a99baca522db8c2ef72bbf3ac',
-    action: 'No',
-    shares: 12.2,
-    date: 1718901111,
-  },
-  {
-    user: '0xf9d7aeaa523906f59070eee225f16c7893e2b262',
-    action: 'Yes',
-    shares: 0.42,
-    date: 1718801111,
-  },
-];
+const getTagColorScheme = (transactionType: TransactionType): TagColorSchemeProp => {
+  switch (transactionType) {
+    case TransactionType.Buy:
+      return 'success';
+    case TransactionType.Sell:
+      return 'danger';
+    case TransactionType.Add:
+      return 'info';
+    case TransactionType.Remove:
+      return 'quaternary';
+    default:
+      return 'info';
+  }
+};
 
-export const ActivityTable = () => {
+export const ActivityTable = ({ id }: { id: string }) => {
+  const { data, isLoading } = useQuery({
+    queryKey: ['getMarketTransactions', id],
+    queryFn: async () =>
+      getMarketTransactions({
+        first: 20,
+        id: id.toLowerCase(),
+      }),
+  });
+
+  const activities = data?.fpmmTransactions ?? [];
+
   return (
     <Table>
       <TableHeader>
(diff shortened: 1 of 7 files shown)


app/markets/ActivityTable.tsx
```diff
@@ -6,81 +6,104 @@ import {
...
+      case 'buy':
+        return 'success';
+      case 'sell':
+        return 'danger';
+      case 'add':
+        return 'info';
+      default:
+        return 'info';
+    }
+  };
```
berteotti, line 37: Things that we can improve:

- Call it `transactionType` instead of `option`;
- Use `TransactionType` type in the option instead of `string`;
- No need for lower casing;
- Use `TransactionType.Buy`, etc. inside each case clause;
- Declare this function outside the component.

app/markets/ActivityTable.tsx
```diff
@@ -6,81 +6,104 @@ import {
...
+                  size="xs"
+                  colorScheme={getTagColorScheme(activity.transactionType)}
+                  className="w-fit uppercase"
+                >
+                  {activity.transactionType}
+                </Tag>
+              </TableCell>
+              <TableCell className="truncate text-text-high-em">
+                <p className="w-24 truncate" title={activity.sharesOrPoolTokenAmount}>
+                  {activity.sharesOrPoolTokenAmount}
```
berteotti, line 76: Let's use `formatEther`here.

app/markets/ActivityTable.tsx
```diff
@@ -6,81 +6,104 @@ import {
...
+                <a
+                  href={gnosisExplorerAddress(activity.user.id)}
+                  className="hover:underline"
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  {shortenAddress(activity.user.id)}
+                </a>
+              </TableCell>
+              <TableCell className="">
```
berteotti, line 65: No className needed

app/markets/HistorySection.tsx
```diff
@@ -1,7 +1,7 @@
 import { ActivityTable } from '@/app/markets/ActivityTable';
 import { Icon, TabBody, TabGroup, TabHeader, TabPanel, TabStyled } from '@swapr/ui';
 
-export const HistorySection = () => {
+export const HistorySection = ({ address }: { address: string }) => {
```
berteotti, line 4: Let's call it id to follow how we declaring it above.

app/markets/MarketDetails.tsx
```diff
@@ -89,7 +89,7 @@ export const MarketDetails = ({ id }: MarketDetailsProps) => {
               )}
             </div>
           )}
-          {tab === 'history' && <HistorySection />}
+          {tab === 'history' && <HistorySection address={id} />}
```
berteotti, line 92: Let's call it id to follow how we declaring it above.

app/markets/ActivityTable.tsx
Diogomartf: Yes, good suggestions, thanks.

berteotti (approved):

app/markets/ActivityTable.tsx
```diff
@@ -6,81 +6,109 @@ import {
...
-              {formatDateTime(activity.date)}
-            </TableCell>
-          </TableRow>
-        ))}
-      </TableBody>
+      {isLoading ? (
+        <LoadingSkeleton />
+      ) : (
+        <TableBody className="text-base font-semibold">
+          {activities.map(activity => (
```
ElRodrigote, line 60: If `activites` can be an empty string, I'd check for `activities.length` just to know if we should display an empty state or not

```suggestion
          {activities.length && activities.map(activity => (
```

app/markets/ActivityTable.tsx
```diff
@@ -6,81 +6,109 @@ import {
...
+            <TableRow key={activity.transactionHash}>
+              <TableCell className="text-text-high-em">
+                <a
+                  href={gnosisExplorerAddress(activity.user.id)}
+                  className="hover:underline"
+                  target="_blank"
+                  rel="noopener noreferrer"
```
ElRodrigote, line 67: Love it !

app/markets/ActivityTable.tsx
```diff
@@ -6,81 +6,109 @@ import {
...
 };
+
+const LoadingSkeleton = () => (
+  <TableBody className="text-base font-semibold">
+    {[1, 2, 3, 4, 5, 6, 7, 8].map(fakeActivity => (
+      <TableRow key={fakeActivity}>
+        <TableCell className="text-text-high-em">
+          <div className="h-[18px] w-[90px] animate-pulse rounded-8 bg-outline-low-em"></div>
+        </TableCell>
+        <TableCell className="">
```
ElRodrigote, line 102: Do we need this className or it's a typo and we can remove it?

app/markets/HistorySection.tsx
```diff
@@ -1,7 +1,7 @@
 import { ActivityTable } from '@/app/markets/ActivityTable';
 import { Icon, TabBody, TabGroup, TabHeader, TabPanel, TabStyled } from '@swapr/ui';
 
-export const HistorySection = () => {
+export const HistorySection = ({ id }: { id: string }) => {
   return (
     <div className="p-2">
```
ElRodrigote, line 6: I know it's outside your changes, but maybe we can drop this `div` element and move the `className` into `<TabGroup>`

utils/token.ts
```diff
@@ -1,2 +1,5 @@
 export const shortenAddress = (address: string) =>
   `${address.slice(0, 6)}...${address.slice(-4)}`;
+
+export const gnosisExplorerAddress = (address: string) =>
```
ElRodrigote, line 4: Minor, non-blocking:
This naming feels weird to me, it's a function but I'd expect maybe a string from this variable name.

WDYT about renaming it into a verb, like `exploreGnosisAddress` or something like that?

app/markets/ActivityTable.tsx
Diogomartf: typo

app/markets/HistorySection.tsx
Diogomartf: `TabGroup` is `Fragment`

utils/token.ts
Diogomartf: I'll take it.

berteotti: I would call it `getGnosisAddressExplorerLink` instead

app/markets/ActivityTable.tsx
ElRodrigote: Sorry my bad, meant an empty array

Diogomartf: Oh the empty state, not very important right now. We can handle it after if see the need.

PR #57, opened 2024-07-03
[PGIO-47][FEAT] - Improve markets outcome bar

ElRodrigote:
## Description
* Update the OutcomeBar component to use current market data if available, or fetch the last known trade data to draw the bar
* Improves the loading status handling for the outcome bar if needed

## Visual Evidence
https://www.loom.com/share/996aefa3be284e2caa5d0dec994badf2?sid=5144930a-df8f-4f5b-9f58-3f000b9c7a95

### DIFF
--- a/app/components/OutcomeBar.tsx
+++ b/app/components/OutcomeBar.tsx
@@ -1,14 +1,15 @@
+import { cx } from 'class-variance-authority';
 import { useQuery } from '@tanstack/react-query';
+import request from 'graphql-request';
+
+import { BlockDataType } from '@/queries/xdai';
 import {
   FixedProductMarketMaker,
   FpmmTrade_OrderBy,
   getMarketTrades,
 } from '@/queries/omen';
-import { cx } from 'class-variance-authority';
-import { Outcome } from '@/entities';
-import request from 'graphql-request';
 import { OMEN_SUBGRAPH_URL, XDAI_BLOCKS_SUBGRAPH_URL } from '@/constants';
-import { BlockDataType } from '@/queries/xdai';
+import { Outcome } from '@/entities';
 
 interface OutcomeBarProps {
   market: FixedProductMarketMaker;
--- a/app/components/OutcomeBar.tsx
+++ b/app/components/OutcomeBar.tsx
@@ -84,11 +84,11 @@ export const OutcomeBar = ({ market }: OutcomeBarProps) => {
     market.outcomeTokenMarginalPrices?.[1] ?? lastTradeMarginalPrices?.[1]
   );
 
-  const hasPercentages = outcome0.percentage && outcome1.percentage;
+  const outcomeHasPercentages = outcome0.percentage && outcome1.percentage;
 
   return (
     <div className="space-y-1">
-      {!hasPercentages ? (
+      {!outcomeHasPercentages ? (
         <ShimmerBar />
       ) : (
         <div className="flex space-x-1">
@@ -119,7 +119,7 @@ export const OutcomeBar = ({ market }: OutcomeBarProps) => {
         </div>
       )}
 
-      {hasPercentages && (
+      {outcomeHasPercentages && (
         <div className="flex justify-between text-sm font-semibold">
           <p className="w-full uppercase text-text-success-main">{`${outcome0.symbol} ${outcome0.percentage || '-'}%`}</p>
           <p className="w-full text-right uppercase text-text-danger-main">
--- a/app/components/OutcomeBar.tsx
+++ b/app/components/OutcomeBar.tsx
@@ -11,7 +11,7 @@ import {
 import { OMEN_SUBGRAPH_URL, XDAI_BLOCKS_SUBGRAPH_URL } from '@/constants';
 import { Outcome } from '@/entities';
 
-const LAST_TRADE_AMOUNT = 1;
+const MIN_TRADES_AMOUNT = 1;
 const blockNumberQuery = `
   query blockNumberByTimestamp($lastTradeTimestamp: String) {
     blocks(where: {timestamp: $lastTradeTimestamp}, first: 1) {
@@ -31,10 +31,9 @@ interface OutcomeBarProps {
   market: FixedProductMarketMaker;
 }
 
+type BlockNumberResponse = { blocks: { number: string }[] };
 type OutcomeTokenMarginalPricesResponse = {
-  [key: string]: {
-    outcomeTokenMarginalPrices: string[];
-  };
+  fixedProductMarketMaker: { outcomeTokenMarginalPrices: string[] };
 };
 
 export const OutcomeBar = ({ market }: OutcomeBarProps) => {
@@ -44,7 +43,7 @@ export const OutcomeBar = ({ market }: OutcomeBarProps) => {
     queryKey: ['getLastMarketTrade', id],
     queryFn: async () =>
       getMarketTrades({
-        first: LAST_TRADE_AMOUNT,
+        first: MIN_TRADES_AMOUNT,
         fpmm: id,
         orderBy: FpmmTrade_OrderBy.CreationTimestamp,
         orderDirection: OrderDirection.Desc,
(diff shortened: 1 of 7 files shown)


app/components/OutcomeBar.tsx
```diff
@@ -1,64 +1,136 @@
...
+        trades.fpmmTrades[trades.fpmmTrades.length - 1]?.creationTimestamp;
+
+      const blockNumber = await request<BlockDataType>(
+        XDAI_BLOCKS_SUBGRAPH_URL,
+        `query blockNumberByTimestamp {
+          _${lastTradeTimestamp}: blocks(where: {timestamp: ${lastTradeTimestamp}}, first: 1) {
+              number
+            }
+        }`
+      );
```
Diogomartf, line 51: are we using this query elsewhere, if so we should probably extract it, otherwise it's okay.

app/components/OutcomeBar.tsx
```diff
@@ -1,64 +1,136 @@
...
     1,
     market.outcomes?.[1] || 'Option 2',
-    market.id,
-    market.outcomeTokenMarginalPrices?.[1],
-    market.question?.currentAnswer
+    id,
+    market.outcomeTokenMarginalPrices?.[1] ?? lastTradeMarginalPrices?.[1]
   );
 
+  const hasPercentages = outcome0.percentage && outcome1.percentage;
```
Diogomartf, line 87: is there outcomes without percentage or are is it always a temporary state?

Diogomartf:
Looking very good 🔥

app/components/OutcomeBar.tsx
```diff
@@ -1,64 +1,136 @@
...
 export const OutcomeBar = ({ market }: OutcomeBarProps) => {
+  const { id } = market;
+
+  const { data: trades } = useQuery({
+    queryKey: ['getMarketTrades', id],
+    queryFn: async () =>
+      getMarketTrades({
+        fpmm: id,
+        orderBy: FpmmTrade_OrderBy.CreationTimestamp,
+      }),
```
berteotti, line 33: Let's fetch only one (the last) market trade.
If it's not the last one to be returned, use a different order direction.

```suggestion
    queryKey: ['getLastMarketTrade', id],
    queryFn: async () =>
      getMarketTrades({
        first: 1,
        fpmm: id,
        orderBy: FpmmTrade_OrderBy.CreationTimestamp,
      }),
```

app/components/OutcomeBar.tsx
```diff
@@ -1,64 +1,136 @@
...
+  });
+
+  const { data: lastTradeMarginalPrices } = useQuery({
+    queryKey: ['getLastTradeMarginalPrices', id],
+    queryFn: async (): Promise<string[] | undefined> => {
+      if (!trades) return;
+
+      const lastTradeTimestamp =
+        trades.fpmmTrades[trades.fpmmTrades.length - 1]?.creationTimestamp;
```
berteotti, line 42: Let's try to have something like this. An array with 1 entry only.

```suggestion
      const lastTradeTimestamp =
        trades.fpmmTrades[0]?.creationTimestamp;
```

app/components/OutcomeBar.tsx
berteotti, line 51: Let's try to have something like this. The query can even go outside. `BlockDataType` will be different.

```suggestion
      const blockNumber = await request<BlockDataType>(
        XDAI_BLOCKS_SUBGRAPH_URL,
        "query blockNumberByTimestamp({ $lastTradeTimestamp: Int }) {
            blocks(where: {timestamp: $lastTradeTimestamp}, first: 1) {
              number
            }
        }",
        { lastTradeTimestamp }
      );
```

app/components/OutcomeBar.tsx
```diff
@@ -1,64 +1,136 @@
...
+        OMEN_SUBGRAPH_URL,
+        `query marginalPricesByBlockNumber($id: ID!) {
+         _${tradeBlockNumber}: fixedProductMarketMaker(id: $id, block: { number: ${tradeBlockNumber} }) {
+              outcomeTokenMarginalPrices
+            }
+        }`,
+        {
+          id,
+        }
+      );
```
berteotti, line 67: Similar to the fix above. `OutcomeTokenMarginalPricesResponse` will be different.

```suggestion
      const marginalPricesResponse = await request<OutcomeTokenMarginalPricesResponse>(
        OMEN_SUBGRAPH_URL,
        `query marginalPricesByBlockNumber($id: ID!, $tradeBlockNumber: Int) {
            fixedProductMarketMaker(id: $id, block: { number: $tradeBlockNumber }) {
              outcomeTokenMarginalPrices
            }
        }`,
        {
          id,
          tradeBlockNumber
        }
      );
```

queries/xdai/types.ts
```diff
@@ -0,0 +1 @@
+export type BlockDataType = Record<string, { number: string }[]>;
```
berteotti, line 1: I think we shouldn't export this from here. It's not going to be used outside `ActivityChart` for now. Types will be different in both solutions.

app/components/OutcomeBar.tsx
ElRodrigote: The `ShimmerBar` is more like a loading state in general. I called it "Shimmer" due to the effect (it's not a spinner animation or a loading animation), I don't mind renaming it as "LoadingBar" or something like that if it's easier to read and understand.

We should always get a percentage response. We'll try to get it out the `market` data first, and if there's no market percentages data, we try to get it from the subgraph (first from XDAI explorer to get the block number, then from Omen with that block number).

I tried to use the `fetching` value out the `lastTrade` query data, the thing is that for many states we're using the `market` data and not the `lastTrade`, so we would force a not needed loading state (because we already have that data).

That's why I decided to use these two variables percentages, in a sense of "ok let's the outcome class handle the decision and we'll use the fastest response from that class"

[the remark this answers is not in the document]
ElRodrigote: Ahhh I decided to not name it like that because we were inside an Outcome component, but you're right maybe it's easier to read and to get a result when you search for "outcome" variables in the codebase, thanks!

app/components/OutcomeBar.tsx
ElRodrigote: Thanks for the suggestion! Renamed the variable and fixed the `getMarketTrades` fn to accept the `first` parameter!

app/components/OutcomeBar.tsx
ElRodrigote: Agree! Updating soon!

app/components/OutcomeBar.tsx
ElRodrigote: Pointed! Working on it, thanks!

app/components/OutcomeBar.tsx
ElRodrigote: Noted! Doing it

queries/xdai/types.ts
ElRodrigote: Yeah with the suggested changes you've made I already removed this from here <3

app/components/OutcomeBar.tsx
```diff
@@ -1,64 +1,145 @@
...
+  });
+
+  const { data: lastTradeMarginalPrices } = useQuery({
+    queryKey: ['getLastTradeMarginalPrices', id],
+    queryFn: async (): Promise<string[] | undefined> => {
+      if (!trade) return;
+
+      const lastTradeTimestamp = trade.fpmmTrades[0]?.creationTimestamp;
+
+      const blockNumber = await request<{ blocks: { number: string }[] }>(
```
berteotti, line 61: Let's give a name to this type

berteotti (approved):
Good job Rorry!

app/components/OutcomeBar.tsx
Diogomartf: Renaming `ShimmerBar` to `LoadingOutcomeBar`, makes it clear on what it is. 

I think It's not that relevant what kind of loading it is, if a spinner or shimmer but that is a loading.

Related to the `hasOutcomePercentages`, I'm wondering if would be more clear if we did something like this

```js
const isLoading = !hasPercentages
```
And the we use the `isLoading` variable

Diogomartf (approved):

PR #58, opened 2024-07-03
feat: add pagination to activity table

berteotti:
image

### DIFF
--- a/app/markets/ActivityTable.tsx
+++ b/app/markets/ActivityTable.tsx
@@ -13,8 +13,10 @@ import {
   getGnosisAddressExplorerLink,
   shortenAddress,
 } from '@/utils';
-import { Tag, TagColorSchemeProp } from '@swapr/ui';
+import { Button, Icon, Tag, TagColorSchemeProp } from '@swapr/ui';
 import { useQuery } from '@tanstack/react-query';
+import { cx } from 'class-variance-authority';
+import { useState } from 'react';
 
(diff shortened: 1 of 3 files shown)


Diogomartf:
🔥 Looking aweasomeeeee! 🔥

**Just one design detail**
When changing pages there is a little dance probably because the elements change width, would be great to make that fixed.

Eg:
https://a533d3ac.presagio.pages.dev/markets?id=0x3f16a2ba3e173852f9a8742174bf4c3d6b39a58b

app/markets/ActivityTable.tsx
```diff
@@ -31,64 +33,107 @@ const getTagColorScheme = (transactionType: TransactionType): TagColorSchemeProp
...
       getMarketTransactions({
-        first: 20,
+        first: ITEMS_PER_PAGE,
+        skip: page * ITEMS_PER_PAGE,
         id: id.toLowerCase(),
       }),
   });
 
+  const hasMoreMarkets =
+    transactionsNextPage && transactionsNextPage.fpmmTransactions.length !== 0;
```
ElRodrigote, line 62: Minor, non-blocking

```suggestion
    transactionsNextPage && transactionsNextPage.fpmmTransactions.length;
```

ElRodrigote (approved):

app/markets/ActivityTable.tsx
berteotti: I find `!==0` more readable. That way I can easily understand that it isn't an empty array.

Diogomartf (approved):

PR #59, opened 2024-07-04
feat: add thumbnail image

berteotti:
image


Diogomartf:
image

only some markets have an image? :(

app/components/Thumbnail.tsx
Diogomartf, line 1: `Thumbnail` is connected to IPFS, so it should be in the naming.

app/components/CardBet.tsx
```diff
@@ -141,7 +142,12 @@ export const CardBet = ({ userPosition }: BetProps) => {
             </p>
           </div>
           <div className="flex space-x-4">
-            <div className="size-[40px] rounded-8 bg-text-low-em bg-gradient-to-r from-[#cb8fc1] to-[#b459c6]" />
+            <Thumbnail
+              width={40}
+              height={40}
+              className="size-[40px] rounded-8"
```
ElRodrigote, line 148: Minor, non-blocking

Maybe we can abstract the `40` into a const and reuse, something like `THUMBNAIL_SIZE`

utils/ipfs.ts
```diff
@@ -0,0 +1,12 @@
+import base58 from 'bs58';
+
+export function byte32ToIPFSCIDV0(hexstr: string): string {
+  /**
+   * Convert 32 bytes hex string to ipfscidv0.
+   * @param hexstr - 32 Bytes long string
+   * @returns IPFS CID Version 0
+   */
+  const binaryStr = Buffer.from(hexstr, 'hex');
```
ElRodrigote, line 9: Are we sure we will always get a correct encoded and value for the image hash from Omen?

In my experience using buffers, they can become tricky with edge cases. I saw you're checking for the image hash to exist as a string, just pondering on the formats and encoding

Diogomartf (approved):

utils/ipfs.ts
berteotti: Yes, we will always get a good value. Otherwise the image will fails, which isn't a big deal.

app/components/CardBet.tsx
berteotti: Will leave it as it is for now

PR #60, opened 2024-07-04
feat: improve go back button

berteotti:
This is just an experience. I feel like this button is weird to interact. I made it a bit less interactive so it feels cleaner. 

image


Diogomartf:
I would make the Back button more like this:

image

We can remove the text, that would make it simpler.

PR #62, opened 2024-07-08
[FIX] - OutcomeBar not loading correctly

ElRodrigote:
## Description
* `isLoading` variable was not refreshing correctly

### DIFF
--- a/app/components/OutcomeBar.tsx
+++ b/app/components/OutcomeBar.tsx
@@ -39,7 +39,7 @@ type OutcomeTokenMarginalPricesResponse = {
 export const OutcomeBar = ({ market }: OutcomeBarProps) => {
   const { id } = market;
 
-  const { data: trade } = useQuery({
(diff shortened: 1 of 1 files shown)


berteotti (commented):
Let's use useQuery `isLoading` to handle the loading state

app/components/OutcomeBar.tsx
```diff
@@ -37,6 +39,8 @@ type OutcomeTokenMarginalPricesResponse = {
 };
 
 export const OutcomeBar = ({ market }: OutcomeBarProps) => {
+  const [isLoading, setIsLoading] = useState(false);
+  const [hasOutcomePercentages, setHasOutcomePercentages] = useState(false);
   const { id } = market;
 
   const { data: trade } = useQuery({
```
Diogomartf, line 46: can you check if we can use `isLoading` from react query?

```suggestion
  const { data: trade, isLoading } = useQuery({
```

berteotti (approved):

PR #63, opened 2024-07-10
[PGIO-53][FEAT] - Improve actions in the Market activity table

ElRodrigote:
## Description
* Updates the market `ActivityTable` component to display the traded outcome as an activity action instead displaying the transaction type.

## Visual Evidence
Screenshot 2024-07-10 at 12 08 20

### DIFF
--- a/app/markets/ActivityTable.tsx
+++ b/app/markets/ActivityTable.tsx
@@ -19,6 +19,7 @@ import {
   shortenAddress,
 } from '@/utils';
 import { Button, Icon, Tag, TagColorSchemeProp } from '@swapr/ui';
+import { Outcome } from '@/entities';
 
 const TAG_COLOR_SCHEMES: { 0: TagColorSchemeProp; 1: TagColorSchemeProp } = {
   0: 'success',
@@ -75,9 +76,14 @@ export const ActivityTable = ({ id }: { id: string }) => {
           <TableBody className="text-base font-semibold">
             {activities?.map(activity => {
               const outcomes = activity.fpmm.outcomes;
-              const outcomeIndex: 0 | 1 = activity.outcomeIndex;
 
-              if (!outcomes || !outcomeIndex) return null;
+              const outcome = new Outcome(
+                activity.outcomeIndex,
+                outcomes?.[activity.outcomeIndex] ?? 'Option 1',
+                id
+              );
+
+              if (!outcomes || !outcome.index) return null;
 
               return (
                 <TableRow key={activity.transactionHash}>
@@ -94,10 +100,10 @@ export const ActivityTable = ({ id }: { id: string }) => {
                   <TableCell>
                     <Tag
                       size="xs"
-                      colorScheme={TAG_COLOR_SCHEMES[outcomeIndex]}
+                      colorScheme={TAG_COLOR_SCHEMES[outcome.index as 0 | 1]}
                       className="w-fit uppercase"
                     >
-                      {outcomes[outcomeIndex]}
+                      {outcome.symbol}
                     </Tag>
                   </TableCell>
                   <TableCell className="truncate text-text-high-em">
--- a/app/markets/ActivityTable.tsx
+++ b/app/markets/ActivityTable.tsx
@@ -1,3 +1,8 @@
+import { useState } from 'react';
+
+import { cx } from 'class-variance-authority';
+import { useQuery } from '@tanstack/react-query';
+
 import {
   Table,
   TableBody,
@@ -7,10 +12,11 @@ import {
   TableRow,
 } from '@/app/components/ui/Table';
 import {
+  FixedProductMarketMaker,
   FpmmTrade_OrderBy,
   OrderDirection,
+  Scalars,
   getMarketTrades,
-  getMarketTransactions,
 } from '@/queries/omen';
 import {
   formatDateTime,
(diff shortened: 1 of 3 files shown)


app/markets/ActivityTable.tsx
```diff
@@ -18,16 +23,12 @@ import { useQuery } from '@tanstack/react-query';
...
-const getTagColorScheme = (transactionType: TransactionType): TagColorSchemeProp => {
-  switch (transactionType) {
-    case TransactionType.Buy:
+const getTagColorScheme = (outcome: string): TagColorSchemeProp => {
+  switch (outcome) {
+    case 'Yes':
       return 'success';
-    case TransactionType.Sell:
+    case 'No':
       return 'danger';
```
berteotti, line 31: Let's use outcome index here. There are markets where outcome string isn't "yes" or "no"

app/markets/ActivityTable.tsx
```diff
@@ -38,7 +39,7 @@ const ITEMS_PER_PAGE = 10;
 export const ActivityTable = ({ id }: { id: string }) => {
   const [page, setPage] = useState(1);
 
-  const { data, isLoading } = useQuery({
+  const { data: txs, isLoading: isTxsLoading } = useQuery({
     queryKey: ['getMarketTransactions', id, page - 1],
```
berteotti, line 43: Let's delete this query and just make a query to `getMarketTrades`. Let's only iterate through  `trades`

app/markets/ActivityTable.tsx
```diff
@@ -48,67 +49,89 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
-                </TableCell>
-                <TableCell className="truncate text-text-high-em">
-                  {formatEtherWithFixedDecimals(activity.sharesOrPoolTokenAmount)}
-                </TableCell>
-                <TableCell className="text-nowrap pr-4 text-right text-text-low-em">
-                  {formatDateTime(activity.creationTimestamp)}
-                </TableCell>
-              </TableRow>
-            ))}
+            {txsActivities?.map(activity => {
```
berteotti, line 96: Here we should only iterate through trades activities

app/markets/ActivityTable.tsx
ElRodrigote: You're right! I'll take the omen types here. Thanks!

app/markets/ActivityTable.tsx
ElRodrigote: Removed!

app/markets/ActivityTable.tsx
ElRodrigote: Done!

Diogomartf (approved):

app/markets/ActivityTable.tsx
```diff
@@ -6,71 +11,69 @@ import {
...
-      return 'quaternary';
-    default:
-      return 'info';
-  }
+const getTagColorScheme = (
+  outcome: Scalars['String']['output'],
+  outcomes: NonNullable<FixedProductMarketMaker['outcomes']>
+): TagColorSchemeProp => {
+  if (outcome === outcomes[0]) return 'success';
+  else if (outcome === outcomes[1]) return 'danger';
```
berteotti, line 34: let's use outcome Index instead of outcome object here.

app/markets/ActivityTable.tsx
```diff
@@ -80,35 +83,42 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
+                      {shortenAddress(activity.creator.id)}
+                    </a>
+                  </TableCell>
+                  <TableCell>
+                    <Tag
+                      size="xs"
+                      colorScheme={getTagColorScheme(outcomes[outcomeIndex], outcomes)}
+                      className="w-fit uppercase"
+                    >
+                      {outcomes[outcomeIndex]}
```
berteotti, line 110: Let's declare an Outcome instance and use the outcome symbol here. We shorten outcomes so they can fit in components like this.

queries/omen/types.ts
```diff
@@ -2339,6 +2339,7 @@ export type FpmmTransaction = {
   __typename?: 'FpmmTransaction';
   additionalSharesCost?: Maybe<Scalars['BigInt']['output']>;
   collateralTokenAmount: Scalars['BigInt']['output'];
+  outcomes?: Maybe<Array<Scalars['String']['output']>>;
```
berteotti, line 2342: Did you add this manually?

queries/omen/markets.ts
```diff
@@ -250,6 +250,7 @@ const getMarketTransactionsQuery = gql`
       }
       fpmm {
         collateralToken
+        outcomes
```
berteotti, line 253: We can delete this, right?

app/markets/ActivityTable.tsx
berteotti: Let's have an object instead of a function. Outcomes will be between 0 and 1 so we can have an object with `1: "success", ...`.

queries/omen/markets.ts
ElRodrigote: Yes! Thanks for the heads up! Forgot to clean this up after consuming the trades only

queries/omen/types.ts
ElRodrigote: Yes! I copied the type from the trades, but we can delete this as we're using the Trades only. Thanks for the heads up!

app/markets/ActivityTable.tsx
ElRodrigote: Ah, you're right, makes much more sense, implementing this one now!

app/markets/ActivityTable.tsx
ElRodrigote: I didn't fully understand this comment, sorry, could you elaborate a bit on what would you like to see happening here?

app/markets/ActivityTable.tsx
```diff
@@ -80,35 +73,42 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
-                  {formatEtherWithFixedDecimals(activity.sharesOrPoolTokenAmount)}
-                </TableCell>
-                <TableCell className="text-nowrap pr-4 text-right text-text-low-em">
-                  {formatDateTime(activity.creationTimestamp)}
-                </TableCell>
-              </TableRow>
-            ))}
+            {activities?.map(activity => {
+              const outcomes = activity.fpmm.outcomes;
+              const outcomeIndex: 0 | 1 = activity.outcomeIndex;
```
berteotti, line 78: Let's use `Outcome` class here. And let's use its attributes below, like `symbol`.

We just need to instantiate the Outcome class equal to `outcomeIndex`.

app/markets/ActivityTable.tsx
berteotti: Commented here: https://github.com/SwaprHQ/presagio/pull/63/files#r1677534154

berteotti (approved):
Looking good!

PR #64, opened 2024-07-15
feat: label better as ai

berteotti:
Fetching AI agents addresses using Dune API

image


app/markets/ActivityTable.tsx
```diff
@@ -55,6 +61,29 @@ export const ActivityTable = ({ id }: { id: string }) => {
       }),
   });
 
+  const { data: aiAgentsList } = useQuery({
+    queryKey: ['getAIAgents'],
+    queryFn: async () => {
+      const options: LatestResultArgs = {
+        queryId: 3582994,
```
ElRodrigote, line 68: Out of curiosity: how do we get this number?

Diogomartf: the answer to that should be on a constant variable so it's more clear where the number comes from:

`const DUNE_QUERY_ID = 3582994`

berteotti: Improved

berteotti: > Out of curiosity: how do we get this number?

It's the agents info query id on dune

app/markets/ActivityTable.tsx
```diff
@@ -96,6 +128,9 @@ export const ActivityTable = ({ id }: { id: string }) => {
                     >
                       {shortenAddress(activity.creator.id)}
                     </a>
+                    {getIsAIAgent(activity.creator.id) && (
+                      <Image src="/ai.svg" alt="ai" width={18} height={18} />
+                    )}
```
ElRodrigote, line 133: WDYT about one of these approaches? The idea is to align a bit the elements in the activity table (UI-wise). I forced the AI tag to display conditionally for these examples purposes only.

AI tag first, conditionally display an empty state if no tag (needs an empty div to fill the column):
```tsx
<TableCell className="grid w-1/3 grid-cols-2 items-center gap-2 pl-4 text-text-high-em">
  {getIsAIAgent(activity.creator.id) && idx % 2 === 0 ? (
    <Image src="/ai.svg" alt="ai" width={18} height={18} />
  ) : (
    <div />
  )}
  <a
    href={getGnosisAddressExplorerLink(activity.creator.id)}
    className="hover:underline"
    target="_blank"
    rel="noopener noreferrer"
  >
    {shortenAddress(activity.creator.id)}
  </a>
</TableCell>
```
Screenshot 2024-07-16 at 11 28 46


Current order, but using columns to display the cell elements (doesn't need an empty div):
```tsx
<TableCell className="grid grid-cols-2 items-center gap-2 pl-4 text-text-high-em">
  <a
    href={getGnosisAddressExplorerLink(activity.creator.id)}
    className="hover:underline"
    target="_blank"
    rel="noopener noreferrer"
  >
    {shortenAddress(activity.creator.id)}
  </a>
  {getIsAIAgent(activity.creator.id) && idx % 2 === 0 && (
    <Image src="/ai.svg" alt="ai" width={18} height={18} />
  )}
</TableCell>
```
Screenshot 2024-07-16 at 11 37 30

app/markets/ActivityTable.tsx
```diff
@@ -20,6 +20,12 @@ import {
 } from '@/utils';
 import { Button, Icon, Tag, TagColorSchemeProp } from '@swapr/ui';
 import { Outcome } from '@/entities';
+import { DuneClient, LatestResultArgs, ParameterType } from '@duneanalytics/client-sdk';
+import Image from 'next/image';
+
+const DUNE_API_KEY = process.env.NEXT_PUBLIC_DUNE_API_KEY || '';
```
Diogomartf, line 26: We should probably add this to `constants`

app/markets/ActivityTable.tsx
berteotti: This seems to be tacking a lot of space. Also it breaks a bit on mobile. I think we can take the current solution for now and check if we can improve later.

app/markets/ActivityTable.tsx
```diff
@@ -96,6 +128,9 @@ export const ActivityTable = ({ id }: { id: string }) => {
                     >
                       {shortenAddress(activity.creator.id)}
                     </a>
+                    {getIsAIAgent(activity.creator.id) && (
+                      <Image src="/ai.svg" alt="ai" width={18} height={18} />
```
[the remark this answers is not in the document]
berteotti: The current AI badge on figma is a SVG. It doesn't quite fit on any button / text style that we currently have. We can ask for new designs but I would update it on another iteration.

Also looks good on light and dark mode.

If we need to have it as text, I can change it.

app/markets/ActivityTable.tsx
berteotti: But otherwise it looks good, I like the tabular style.

app/markets/ActivityTable.tsx
berteotti: Will move it. But not sure about moving constants used only in 1 place to a different file and export it there.

app/markets/ActivityTable.tsx
```diff
@@ -96,6 +127,9 @@ export const ActivityTable = ({ id }: { id: string }) => {
                     >
                       {shortenAddress(activity.creator.id)}
                     </a>
+                    {getIsAIAgent(activity.creator.id) && (
```
Diogomartf, line 130: why the get in the name? 

I think this would make it more clear:

```suggestion
                    {isAIAgent(activity.creator.id) && (
```

app/markets/ActivityTable.tsx
Diogomartf: "AI" on the left doesn't work in this case @ElRodrigote cause it's something that's only present in some cases, not always, in that case right is more appropriate otherwise you'll get a bunch of empty space.

Diogomartf (approved):

PR #66, opened 2024-07-22
feat: enable market collateral token

berteotti:
Enable any market collateral token to be shown and traded on the market.

image


app/components/ui/TokenLogo.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { WXDAI } from '@/constants';
+import { XDAI_LOGO, DEFAULT_TOKEN_LOGO } from '@/public/assets';
+
+const localTokenAssets = [{ address: WXDAI.address, asset: XDAI_LOGO.src }];
+
+const fetchImage = async (url: string) => {
+  const result = await fetch(url);
+
+  if (result.status === 200) return url;
+  else throw null;
```
ElRodrigote, line 15: Mmm, can we handle this error more descriptively? It feels weird to throw `null` on error.

app/components/ui/TokenLogo.tsx
```diff
@@ -0,0 +1,64 @@
...
+                `https://raw.githubusercontent.com/1Hive/default-token-list/master/src/assets/gnosis/${address}/logo.png`
+              ),
+            staleTime: Infinity,
+          })
+          .then(source => setSource(source))
+          .catch(() => setSource(DEFAULT_TOKEN_LOGO.src))
+      );
+  }, [address, queryClient]);
+
+  if (!source) return;
```
ElRodrigote, line 61: Should we get some logo placeholder as we do in our other apps? Like displaying the first token letters in a circle or something like that?

ElRodrigote (approved):

public/assets/question-mark.png
Diogomartf, line 1: Default logo should not be a question mark.

We should show 2 letters of token symbol with bg color for eg.

app/components/ui/TokenLogo.tsx
```diff
@@ -0,0 +1,64 @@
...
+          .fetchQuery({
+            queryKey: ['get1HiveImage', address],
+            queryFn: async () =>
+              await fetchImage(
+                `https://raw.githubusercontent.com/1Hive/default-token-list/master/src/assets/gnosis/${address}/logo.png`
+              ),
+            staleTime: Infinity,
+          })
+          .then(source => setSource(source))
+          .catch(() => setSource(DEFAULT_TOKEN_LOGO.src))
```
Diogomartf, line 57: would be great if we could simplify this bit, but okay as it is for now.

app/components/ui/TokenLogo.tsx
berteotti: The thing is, nothing is done with this error.  But will add a message just in case.

app/components/TokenLogo.tsx
```diff
@@ -0,0 +1,101 @@
...
+    );
+
+  if (source === null)
+    return (
+      <div className={cx(logoStyles({ size }), 'text-[6px] font-bold text-text-black')}>
+        {symbol}
+      </div>
+    );
+
+  return <Logo src={source} alt="token logo" size={size} className={className} />;
```
Diogomartf, line 100:
```suggestion
  return <Logo src={source} alt={`${token.name} token logo`} size={size} className={className} />;
```

Diogomartf (approved):

PR #67, opened 2024-07-24
Improve Market Activity show Add/Remove liquidity and type od trades …

Diogomartf:
…(buy or sell) outcome shares

- join data from trades and transactions with mergeMarketTradesTransactionsArrays
- getMarketTradesAndTransactions subgraph querty
- formatDateTimeWithYear date utils fn

**Notice**
Will update design in another PR to keep this PR simple.

**Screenshot**
image

### DIFF
--- a/app/markets/ActivityTable.tsx
+++ b/app/markets/ActivityTable.tsx
@@ -11,9 +11,18 @@ import {
   TableHeader,
   TableRow,
 } from '@/app/components/ui/Table';
-import { FpmmTrade_OrderBy, OrderDirection, getMarketTrades } from '@/queries/omen';
+import {
+  FpmmTrade,
+  FpmmTrade_OrderBy,
+  FpmmTransaction,
+  FpmmType,
+  OrderDirection,
+  getMarketTrades,
+  getMarketTradesAndTransactions,
+} from '@/queries/omen';
 import {
   formatDateTime,
+  formatDateTimeWithYear,
   formatEtherWithFixedDecimals,
   getGnosisAddressExplorerLink,
   shortenAddress,
@@ -33,32 +42,48 @@ const TAG_COLOR_SCHEMES: { 0: TagColorSchemeProp; 1: TagColorSchemeProp } = {
 
 const ITEMS_PER_PAGE = 10;
 
+type MergedTradeTransaction = Omit<FpmmTrade, '__typename'> &
+  Partial<Omit<FpmmTransaction, '__typename'>> & {
+    __typename?: 'FpmmTrade' | 'FpmmTransaction';
+  };
+
+function mergeMarketTradesAndTransactionsArrays(
+  trades: FpmmTrade[],
+  transactions: FpmmTransaction[]
+): MergedTradeTransaction[] {
+  const tradesAndTransactions = new Map<string, MergedTradeTransaction>();
+
+  trades.forEach(trade => {
+    tradesAndTransactions.set(trade.id, { ...trade, __typename: 'FpmmTrade' });
+  });
+
+  transactions.forEach(transaction => {
+    if (tradesAndTransactions.has(transaction.id)) {
+      const trade = tradesAndTransactions.get(transaction.id)!;
+      tradesAndTransactions.set(transaction.id, {
+        ...trade,
+        ...transaction,
+        fpmm: trade.fpmm,
+        collateralToken: transaction.fpmm?.collateralToken,
+        __typename: 'FpmmTrade',
+      });
+    } else if (transaction.fpmmType === FpmmType.Liquidity) {
+      tradesAndTransactions.set(transaction.id, {
+        ...transaction,
+        collateralToken: transaction.fpmm?.collateralToken,
+        __typename: 'FpmmTransaction',
+      } as MergedTradeTransaction);
+    }
+  });
+
+  // sort tx's and trades
+  return Array.from(tradesAndTransactions.values()).sort((a, b) => {
+    return parseInt(b.creationTimestamp) - parseInt(a.creationTimestamp);
+  });
+}
+
 export const ActivityTable = ({ id }: { id: string }) => {
   const [page, setPage] = useState(1);
-
-  const { data: trades, isLoading } = useQuery({
-    queryKey: ['getMarketTrades', id, page - 1],
-    queryFn: async () =>
-      getMarketTrades({
-        first: ITEMS_PER_PAGE,
-        skip: (page - 1) * ITEMS_PER_PAGE,
-        fpmm: id,
-        orderBy: FpmmTrade_OrderBy.CreationTimestamp,
-        orderDirection: OrderDirection.Desc,
-      }),
-  });
-
-  const { data: tradesNextPage } = useQuery({
-    queryKey: ['getMarketTrades', id, page],
-    queryFn: async () =>
-      getMarketTrades({
-        first: ITEMS_PER_PAGE,
-        skip: page * ITEMS_PER_PAGE,
-        fpmm: id,
-        orderBy: FpmmTrade_OrderBy.CreationTimestamp,
-        orderDirection: OrderDirection.Desc,
-      }),
-  });
 
   const { data: aiAgentsList } = useQuery({
     queryKey: ['getAIAgents'],
@@ -85,9 +110,50 @@ export const ActivityTable = ({ id }: { id: string }) => {
     );
   };
 
+  const { data: marketTradesTransactions, isLoading } = useQuery({
+    queryKey: ['getMarketTradesAndTransactions', id, page - 1],
+    queryFn: async () =>
+      getMarketTradesAndTransactions({
+        first: ITEMS_PER_PAGE,
+        skip: (page - 1) * ITEMS_PER_PAGE,
+        fpmm: id,
+        orderBy: FpmmTrade_OrderBy.CreationTimestamp,
+        orderDirection: OrderDirection.Desc,
+      }),
+  });
+
+  const isMarketTradesTransactions =
+    marketTradesTransactions?.fpmmTrades && marketTradesTransactions?.fpmmTransactions;
+
+  const { data: tradesNextPage } = useQuery({
+    queryKey: ['getMarketTrades', id, page],
+    queryFn: async () =>
+      getMarketTrades({
+        first: ITEMS_PER_PAGE,
+        skip: page * ITEMS_PER_PAGE,
+        fpmm: id,
+        orderBy: FpmmTrade_OrderBy.CreationTimestamp,
+        orderDirection: OrderDirection.Desc,
+      }),
+  });
+
   const hasMoreMarkets = tradesNextPage && tradesNextPage.fpmmTrades.length !== 0;
   const showPaginationButtons = hasMoreMarkets || page !== 1;
-  const activities = trades?.fpmmTrades;
+
+  const marketTradesTransactionsMerged: MergedTradeTransaction[] =
+    isMarketTradesTransactions
+      ? mergeMarketTradesAndTransactionsArrays(
+          marketTradesTransactions.fpmmTrades,
+          marketTradesTransactions.fpmmTransactions
+        )
+      : [];
+
+  const sortedActivities = (activities: MergedTradeTransaction[]) =>
+    activities.sort((a, b) => {
+      return parseInt(b.creationTimestamp) - parseInt(a.creationTimestamp);
+    });
+
+  const activities = sortedActivities(marketTradesTransactionsMerged);
 
   return (
     <div>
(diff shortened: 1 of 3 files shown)


queries/omen/markets.ts
```diff
@@ -332,6 +393,26 @@ const getMarketUserTrades = async (params: QueryFpmmTradesArgs & FpmmTrade_Filte
...
+const getMarketTradesAndTransactions = async (
+  params: QueryFpmmTradesArgs & QueryFpmmTransactionsArgs & FpmmTrade_Filter
+) => {
+  const modifiedParams = {
+    ...params,
+    tradesFirst: params.first || 0,
+    // we need to get 2 more trasanction than trade
+    // to handle liquidity tx (add/remove) that have no associated trade
+    // by getting 2 more tx we match all trades
+    // even if there is 2 liquidity tx in there
```
berteotti, line 405: Can we simplify this message? Something like "Txs can have +2 results than trades. We need to fetch +2 to be able to merge tx and trades"

app/markets/ActivityTable.tsx
```diff
@@ -33,33 +43,54 @@ const TAG_COLOR_SCHEMES: { 0: TagColorSchemeProp; 1: TagColorSchemeProp } = {
...
+//    creator, fpmm, id, outcomeIndex, outcomeTokensTraded
+
+// fields only on transaction
+//    additionalSharesCost, collateralTokenAmount, fpmmType, sharesOrPoolTokenAmount, transactionType, user, __typename
+
+function mergeMarketTradesTransactionsArrays(
+  trades: FpmmTrade[],
+  transactions: FpmmTransaction[]
+): (FpmmTrade & Partial<FpmmTransaction>)[] {
+  const tradesAndTransactions = new Map<string, FpmmTrade & Partial<FpmmTransaction>>();
```
berteotti, line 63: Why are we using a Map? Why not a Set?

app/markets/ActivityTable.tsx
```diff
@@ -33,33 +43,54 @@ const TAG_COLOR_SCHEMES: { 0: TagColorSchemeProp; 1: TagColorSchemeProp } = {
...
+    // if trade has a transaction with same id
+    if (tradesAndTransactions.has(transaction.id)) {
+      const trade = tradesAndTransactions.get(transaction.id)!;
+      tradesAndTransactions.set(transaction.id, {
+        ...trade,
+        ...transaction,
+        fpmm: trade.fpmm,
+        collateralToken: transaction.fpmm?.collateralToken,
+      });
+    } else if (transaction.fpmmType.toLowerCase() === 'liquidity') {
```
berteotti, line 80: You can get 'liquidity' from `transaction.fpmmType` and no need for lower casing.

app/markets/ActivityTable.tsx
```diff
@@ -85,18 +116,62 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
+
+  console.log(
+    'marketTradesTransactionsMerged:',
+    Array.from(marketTradesTransactionsMerged.values())
+  );
+
+  const sortedActivities = marketTradesTransactionsMerged =>
+    Array.from(marketTradesTransactionsMerged.values()).sort((a, b) => {
+      return parseInt(b.creationTimestamp) - parseInt(a.creationTimestamp);
+    });
```
berteotti, line 163: Can't we sort in the subgraph?

app/markets/ActivityTable.tsx
```diff
@@ -85,18 +116,62 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
+
+  const activities = sortedActivities(marketTradesTransactionsMerged);
```
berteotti, line 165: We don't need this function, we can use `Array.from...` directly here

app/markets/ActivityTable.tsx
Diogomartf: Yes. Thanks.

app/markets/ActivityTable.tsx
Diogomartf: Nope, we need to sort it after merge.

app/markets/ActivityTable.tsx
```diff
@@ -108,42 +183,76 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
 
-              if (!outcomes || !outcomeIndex) return null;
+              const hasOutcomes = outcomes && outcomeIndex;
 
-              const outcome = new Outcome(
-                outcomeIndex,
-                outcomes[outcomeIndex] ?? 'Option ' + outcomeIndex,
-                id
-              );
+              const isLiquidityEvent = !hasOutcomes && activity.fpmmType === 'Liquidity';
```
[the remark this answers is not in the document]
Diogomartf: 👍

app/markets/ActivityTable.tsx
Diogomartf: I think the advantage of Set vs Map would be if we need to have unique values, but we only need to ensure key uniqueness and that's a feature on Map.

I think Map suits better.

app/markets/ActivityTable.tsx
```diff
@@ -33,33 +42,49 @@ const TAG_COLOR_SCHEMES: { 0: TagColorSchemeProp; 1: TagColorSchemeProp } = {
...
-      getMarketTrades({
-        first: ITEMS_PER_PAGE,
-        skip: page * ITEMS_PER_PAGE,
-        fpmm: id,
-        orderBy: FpmmTrade_OrderBy.CreationTimestamp,
-        orderDirection: OrderDirection.Desc,
-      }),
+  transactions.forEach(transaction => {
+    if (tradesAndTransactions.has(transaction.id)) {
+      const trade = tradesAndTransactions.get(transaction.id)!;
```
berteotti, line 62:
```suggestion
      const trade = tradesAndTransactions.get(transaction.id);
```

is this a typo?

app/markets/ActivityTable.tsx
```diff
@@ -108,42 +173,81 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
+                              ? TAG_COLOR_SCHEMES[outcome.index as 0 | 1]
+                              : 'quaternary'
+                        }
+                        className="w-fit uppercase"
+                      >
+                        {isLiquidityEvent
+                          ? activity.transactionType
+                          : outcome
+                            ? outcome.symbol
+                            : '??'}
```
berteotti, line 237: Let's just have a dash here. Or unknown.

Diogomartf: Maybe a dash.

berteotti (approved):
Looking good, good job Diogo

app/markets/ActivityTable.tsx
Diogomartf: No. It's used to tell the TypeScript compiler that we're certain there's a value won't be null or undefined, even if TypeScript's type inference suggests it might be.

ElRodrigote (approved):

app/markets/ActivityTable.tsx
```diff
@@ -177,9 +179,165 @@ export const ActivityTable = ({ id }: { id: string }) => {
   );
 };
 
+interface TradeRowProps {
+  activity: MergedTradeTransaction;
+  aiAgentsList: AiAgentsList;
```
berteotti, line 184: Instead of passing down a list, why not pass a bool value to show if is an AI agent? Otherwise we are passing an entire list to every row

berteotti: Or a function to determine if it is

app/markets/ActivityTable.tsx
```diff
@@ -177,9 +179,165 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
+  aiAgentsList,
+}: CreatorAddressLinkWithAILabelProps) => {
+  const getIsAIAgent = useMemo(() => {
+    return (address: string) => {
+      if (!aiAgentsList?.length) return false;
+      return aiAgentsList.some(
+        aiAgent => String(aiAgent.address).toLowerCase() === address.toLowerCase()
+      );
+    };
+  }, [aiAgentsList]);
```
berteotti, line 319: This function can be defined on a higher level

app/markets/ActivityTable.tsx
```diff
@@ -177,9 +179,165 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
+      </TableCell>
+    </TableRow>
+  );
+};
+
+interface CreatorAddressLinkWithAILabelProps {
+  creatorAddress: string;
+  aiAgentsList: AiAgentsList;
+}
+const CreatorAddressLinkWithAILabel = ({
```
berteotti, line 308: This component looks like it's doing too little.

app/markets/ActivityTable.tsx
```diff
@@ -60,94 +97,59 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
-                  </TableCell>
-                </TableRow>
-              );
+              if (tradeAssociatedWithTransaction) {
+                const activity: MergedTradeTransaction = {
+                  ...tradeAssociatedWithTransaction,
+                  ...transaction,
+                  fpmm: tradeAssociatedWithTransaction.fpmm,
+                  collateralToken: transaction.fpmm?.collateralToken,
+                  __typename: 'FpmmTrade',
```
berteotti, line 140: do we need this `_typename`?

app/markets/ActivityTable.tsx
```diff
@@ -60,94 +97,59 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
-                    {formatDateTime(activity.creationTimestamp)}
-                  </TableCell>
-                </TableRow>
-              );
+              if (tradeAssociatedWithTransaction) {
+                const activity: MergedTradeTransaction = {
+                  ...tradeAssociatedWithTransaction,
+                  ...transaction,
+                  fpmm: tradeAssociatedWithTransaction.fpmm,
+                  collateralToken: transaction.fpmm?.collateralToken,
```
berteotti, line 139: Why this? `transaction` already has `collateralToken` and we are doubling it's information

app/markets/ActivityTable.tsx
```diff
@@ -60,94 +97,59 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
-                  <TableCell className="text-nowrap pr-4 text-right text-text-low-em">
-                    {formatDateTime(activity.creationTimestamp)}
-                  </TableCell>
-                </TableRow>
-              );
+              if (tradeAssociatedWithTransaction) {
+                const activity: MergedTradeTransaction = {
+                  ...tradeAssociatedWithTransaction,
+                  ...transaction,
+                  fpmm: tradeAssociatedWithTransaction.fpmm,
```
berteotti, line 138: If we destruct `tradeAssociatedWithTransaction` in last maybe we don't need this line

app/markets/ActivityTable.tsx
Diogomartf: nope

app/markets/ActivityTable.tsx
Diogomartf: bool works.

app/markets/ActivityTable.tsx
Diogomartf: we need, otherwise it keeps `transaction.fpmm` instead of `trade.fpmm` (and we want to keep this one).

app/markets/ActivityTable.tsx
```diff
@@ -177,9 +185,141 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
+
+      <TableCell className="truncate text-text-high-em">
+        <div className="flex items-center space-x-2">
+          <p>
+            {activity.transactionType?.toLowerCase() === 'sell' && (
+              <span className="mr-0.5 text-md font-bold">-</span>
+            )}
+            {activity.transactionType?.toLowerCase() === 'buy' && (
+              <span className="mr-0.5 text-md font-bold">+</span>
+            )}
```
berteotti, line 232: Let use TransactionType here.

```suggestion
            {activity.transactionType?.toLowerCase() === TransactionType.Sell && (
              <span className="mr-0.5 text-md font-bold">-</span>
            )}
            {activity.transactionType?.toLowerCase() === TransactionType.Buy && (
              <span className="mr-0.5 text-md font-bold">+</span>
            )}
```

app/markets/ActivityTable.tsx
berteotti:
```suggestion
                  ...transaction,
                  ...tradeAssociatedWithTransaction,
```

If we do something like this it will keep `trade.fpmm`, tradeAssociatedWithTransaction will override transaction values. Are we doing in that order for any specific reason?

ElRodrigote (approved):

app/markets/ActivityTable.tsx
```diff
@@ -177,9 +185,141 @@ export const ActivityTable = ({ id }: { id: string }) => {
...
+
+      <TableCell className="truncate text-text-high-em">
+        <div className="flex items-center space-x-2">
+          <p>
+            {activity.transactionType?.toLowerCase() === TransactionType.Sell && (
+              <span className="mr-0.5 text-md font-bold">-</span>
+            )}
+            {activity.transactionType?.toLowerCase() === TransactionType.Buy && (
+              <span className="mr-0.5 text-md font-bold">+</span>
+            )}
```
berteotti, line 232: We can't turn it into lower case here otherwise will not work.

Diogomartf: true, fixed.

berteotti (approved):

PR #71, opened 2024-08-05
sort userPositions on My bets page

Diogomartf:
Sorts my bets (userPositions) by position created timestamp.

This lead to a refactor on how we fetch data, we also need to merge all data into an array to be able to sort it properly.

there are still some type issues that I'll fix.

### DIFF
--- a/app/components/CardBet.tsx
+++ b/app/components/CardBet.tsx
@@ -7,7 +7,7 @@ import { useConfig } from 'wagmi';
 import { Button, Tag } from '@swapr/ui';
 import { Card, TokenLogo } from '@/app/components';
 
-import { remainingTime } from '@/utils/dates';
+import { formatDateTime, formatDateTimeWithYear, remainingTime } from '@/utils/dates';
 import {
   Market,
   Position,
@@ -27,7 +27,6 @@ interface BetProps {
 }
 
 export const CardBet = ({ userPositionComplete }: BetProps) => {
-  console.log('userPosition:', userPositionComplete);
   const position = new Position(userPositionComplete.position);
   const outcomeIndex = position.outcomeIndex - 1;
 
@@ -41,6 +40,9 @@ export const CardBet = ({ userPositionComplete }: BetProps) => {
   const collateralAmountUSDSpent = tradesCollateralAmountUSDSpent({
     fpmmTrades: userPositionComplete?.fpmmTrades,
   });
+  const lastTradeTimestamp =
+    userPositionComplete?.fpmmTrades[userPositionComplete?.fpmmTrades.length - 1]
+      .creationTimestamp;
 
   const outcomeBalance = tradesOutcomeBalance({
     fpmmTrades: userPositionComplete?.fpmmTrades,
@@ -124,9 +126,9 @@ export const CardBet = ({ userPositionComplete }: BetProps) => {
           </div>
         </section>
       </Link>
-      <section className="flex h-[56px] items-center border-t border-outline-base-em px-4 md:h-[48px]">
+      <section className="flex items-center border-t border-outline-base-em px-4 py-2 md:h-[48px] md:py-0">
         <div className="flex w-full items-center justify-between space-x-4">
-          <div className="flex flex-col items-start space-y-0.5 md:flex-row md:items-center md:space-x-2">
+          <div className="flex flex-col items-start space-y-0.5 md:flex-row md:items-center md:space-x-3">
             <div className="flex items-center space-x-1">
               <p className="text-sm font-semibold text-text-med-em">Bet amount:</p>
               <p className="text-sm font-semibold text-text-high-em">
@@ -147,6 +149,12 @@ export const CardBet = ({ userPositionComplete }: BetProps) => {
                   '$' + collateralAmountUSDSpent?.toFixed(2)
                 )}
               </div>
+            </div>
+            <div className="flex items-center space-x-1">
+              <p className="text-sm font-semibold text-text-med-em">Last bet:</p>
+              <p className="text-sm font-semibold text-text-high-em">
+                {formatDateTimeWithYear(lastTradeTimestamp)}
+              </p>
             </div>
           </div>
           {canClaim && (
--- a/app/components/CardBet.tsx
+++ b/app/components/CardBet.tsx
@@ -1,15 +1,12 @@
 'use client';
 
-import { useQuery } from '@tanstack/react-query';
 import { cx } from 'class-variance-authority';
 import Link from 'next/link';
-import { useAccount, useConfig } from 'wagmi';
+import { useConfig } from 'wagmi';
 
 import { Button, Tag } from '@swapr/ui';
 import { Card, TokenLogo } from '@/app/components';
 
-import { UserPosition } from '@/queries/conditional-tokens/types';
-import { getConditionMarket, getMarketUserTrades } from '@/queries/omen';
 import { remainingTime } from '@/utils/dates';
 import {
   Market,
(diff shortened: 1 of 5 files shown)


ElRodrigote (approved):

app/my-bets/page.tsx
```diff
@@ -7,96 +7,151 @@ import { Market, Position, tradesOutcomeBalance } from '@/entities';
...
+                id: position.conditionId,
+              });
+              const market = new Market(
+                conditionData.conditions[0]?.fixedProductMarketMakers[0]
+              );
+              return getMarketUserTrades({
+                creator: address.toLowerCase(),
+                fpmm: market.data.id,
+                outcomeIndex_in: [outcomeIndex],
+              });
```
berteotti, line 68: I see we are calling `getConditionMarket` here and above. Why not having a query that calls both `getConditionMarket` and `getMarketUserTrades` and does what we are doing inside the `useEffect`?

app/my-bets/page.tsx
```diff
@@ -7,96 +7,151 @@ import { Market, Position, tradesOutcomeBalance } from '@/entities';
...
+  getConditionMarket,
+  getMarketUserTrades,
+} from '@/queries/omen';
+import { useQuery, useQueries } from '@tanstack/react-query';
+
+import { PropsWithChildren, ReactNode, useEffect, useMemo, useState } from 'react';
 import { TabBody, TabGroup, TabHeader, TabPanel, TabStyled } from '@swapr/ui';
 import { useAccount } from 'wagmi';
 
+interface UserPositionAll extends UserPosition {
```
berteotti, line 23: I think we can improve the name. Adding `All` doesn't give much about the type. What about `UserCompletePositionInfo`, `UserPositionCondition`, `ConditionUserPosition`?

app/my-bets/page.tsx
```diff
@@ -7,96 +7,151 @@ import { Market, Position, tradesOutcomeBalance } from '@/entities';
...
-  const userPositions = data?.userPositions ?? [];
+  const userPositions = useMemo(
+    () => userPositionsData?.userPositions ?? [],
+    [userPositionsData]
+  );
+
+  const queries = useMemo(() => {
+    return userPositions.flatMap(userPosition => {
+      const position = new Position(userPosition.position);
+      const outcomeIndex = position.outcomeIndex - 1;
```
ElRodrigote, line 47: Why are we substracting 1 from the position outcome index?

Diogomartf: That's a good question. I don't have a good answer.

This is a detail implementation of the contracts/subgraphs that we're using but I don't think we know the why they did it this way.

@berteotti thoughts?

app/my-bets/page.tsx
Diogomartf: Yeah, seems like a good suggestion, will to make it work.

ElRodrigote (approved):

app/my-bets/page.tsx
Diogomartf: the idea was having user position and "all" the extra related data, but I agree we can improve.

This is a `UserPosition` with `fpmmTrades`, `condition` and `market` data.

Maybe complete is a good description.

i think `UserPositionComplete` can work

app/my-bets/page.tsx
Diogomartf: updated it

ElRodrigote (approved):

app/components/CardBet.tsx
```diff
@@ -1,16 +1,13 @@
...
-import { useAccount, useConfig } from 'wagmi';
+import { useConfig } from 'wagmi';
 
 import { Button, Tag } from '@swapr/ui';
 import { Card, TokenLogo } from '@/app/components';
 
-import { UserPosition } from '@/queries/conditional-tokens/types';
-import { getConditionMarket, getMarketUserTrades } from '@/queries/omen';
-import { remainingTime } from '@/utils/dates';
+import { formatDateTime, formatDateTimeWithYear, remainingTime } from '@/utils/dates';
```
berteotti, line 10: Unused import

app/my-bets/page.tsx
```diff
@@ -7,96 +7,119 @@ import { Market, Position, tradesOutcomeBalance } from '@/entities';
...
+      const userPositionsData = await getUserPositions({ id: address.toLowerCase() });
+      const userPositions = userPositionsData?.userPositions ?? [];
 
-  const filterActiveBets = userPositions.filter(
-    position => position.position.conditions[0].resolved === false
-  );
-  const filterCompleteBets = userPositions.filter(
-    position => position.position.conditions[0].resolved
-  );
+      const userPositionsComplete = await Promise.all(
```
berteotti, line 40: Should we use `Promise.all` or `Promise.allSettled` ?

app/my-bets/page.tsx
```diff
@@ -7,96 +7,119 @@ import { Market, Position, tradesOutcomeBalance } from '@/entities';
...
   const { address } = useAccount();
-  const [unredeemedBets, setUnredeemedBets] = useState<UserPosition[]>([]);
-  const queryClient = useQueryClient();
 
-  const { data, isLoading } = useQuery({
-    queryKey: ['getUserPositions', address],
-    queryFn: () => getUserPositions({ id: address?.toLowerCase() as string }),
-    enabled: !!address,
-  });
+  const { data: userPositionsComplete, isLoading } = useQuery({
```
berteotti, line 32:
```suggestion
  const { data: userPositionsComplete, isLoading } = useQuery<UserPositionComplete[]>({
```

app/my-bets/page.tsx
```diff
@@ -7,96 +7,119 @@ import { Market, Position, tradesOutcomeBalance } from '@/entities';
...
+        b.fpmmTrades[b.fpmmTrades.length - 1]?.creationTimestamp -
+        a.fpmmTrades[a.fpmmTrades.length - 1]?.creationTimestamp
+      );
+    });
+  };
+
+  const sortedAllBets = sortBetsByNewest(userPositionsComplete ?? []);
+  const sortedCompleteBets = sortBetsByNewest(filterCompleteBets);
+  const sortedActiveBets = sortBetsByNewest(filterActiveBets);
+  const sortedUnredeemedBets = sortBetsByNewest(filterUnredeemedBets);
```
berteotti, line 122: Let's sort before filtering. So we first sort, then we filter by complete, active, unredeemed.

app/my-bets/page.tsx
ElRodrigote: This is a good Q.

So to summarize, IMHO we should use `Promise.allSettled` because if some of the promises fail, the execution will get stopped with our current PR approach. But using `.allSettled` we could handle each response individually, so with this update, we can show the bets that didn't fail, and handle the ones that failed.

ElRodrigote (approved):

app/my-bets/page.tsx
Diogomartf: yes

app/my-bets/page.tsx
```diff
@@ -7,96 +7,134 @@ import { Market, Position, tradesOutcomeBalance } from '@/entities';
...
-  const filterActiveBets = userPositions.filter(
-    position => position.position.conditions[0].resolved === false
+  const sortBetsByNewest = (bets: UserPositionComplete[]) => {
+    return [...bets].sort((a, b) => {
+      return (
+        b.fpmmTrades[b.fpmmTrades.length - 1]?.creationTimestamp -
+        a.fpmmTrades[a.fpmmTrades.length - 1]?.creationTimestamp
+      );
+    });
+  };
```
berteotti, line 94: Do we need this function that is used only once?

berteotti: If we need it can be outside the component. Also we can sort inside the `useQuery`.

Diogomartf: I want this code with a description so it's clear on what it does.

Do you think a variable is better?

Diogomartf: I think a variable is better, as it's only used in one place as you mentioned. But I like it here. Since the useQuery is already doing too much.

berteotti: We can have it outside the component and sort them inside the `useQuery`. We just need what's inside sort.

Diogomartf: Updated it

app/my-bets/page.tsx
Diogomartf: Changed to `allSettled`, although is a bit more complex (cause you have to deal with unfulfilled values), the benefit of showing something instead of nothing if some fail seems worth it.

Updated it.

app/my-bets/page.tsx
```diff
@@ -7,96 +7,131 @@ import { Market, Position, tradesOutcomeBalance } from '@/entities';
...
 import { TabBody, TabGroup, TabHeader, TabPanel, TabStyled } from '@swapr/ui';
 import { useAccount } from 'wagmi';
 
+export interface UserPositionComplete extends UserPosition {
+  fpmmTrades: FpmmTrade[];
+  market: FixedProductMarketMaker;
+  condition: Condition;
+}
+
+const sortByNewestBet = (a: any, b: any) => {
```
berteotti, line 29: We can type it, right?

```suggestion
const sortByNewestBet = (a: UserPositionComplete, b: UserPositionComplete) => {
```

berteotti (approved):
Looking good!

PR #72, opened 2024-08-07
fix: wrong network bug and interface

berteotti:
image


Diogomartf:
design fix:

The network button is bigger than the other items

image

Also we need to think on this "button" cause it ain't a button but looks like it.

app/components/NetworkButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+    !!connectorChainId &&
+    !supportedChains.some(supportedChain => supportedChain.id === connectorChainId);
+
+  return (
+    <div className="hidden select-none items-center justify-center space-x-2 rounded-12 bg-surface-surface-2 p-3 md:flex">
+      <Popover>
+        <PopoverTrigger>
+          <ChainIcon id={connectorChainId || chainId} unsupported={unsupportedNetwork} />
+        </PopoverTrigger>
+        <PopoverContent>Hello</PopoverContent>
```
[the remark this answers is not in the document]
berteotti: Yeah, was just testing. Fixed. Planning to have a tooltip here

berteotti:
> The network button is bigger than the other items
Fixed

> Also we need to think on this "button" cause it ain't a button but looks like it.
True, looks like a button but doesn't behave like one. I think for now it blends well with the other navbar components. But we can improve if there is another way to show it.

app/components/Swapbox.tsx
```diff
@@ -211,6 +214,10 @@ export const Swapbox = ({ market }: { market: FixedProductMarketMaker }) => {
 
   const outcomeList = [outcome0, outcome1];
 
+  const unsupportedNetwork =
+    !!connectorChainId &&
+    !supportedChains.some(supportedChain => supportedChain.id === connectorChainId);
+
```
ElRodrigote, line 220: We could abstract this function because we're also using it in `app/components/NetworkButton.tsx`

ElRodrigote:
This case is weird because I have to change the network from my wallet manually, it would be nice to have the UI prompting me with a user flow similar to the one in Stackly:
If I'm connected to an unsupported network, the UI prompts me with an informative message and displays the supported networks, allowing me to connect quickly and use the app.

In Presagio right now:
Screenshot 2024-08-08 at 11 58 33

In Stackly: 
Screenshot 2024-08-08 at 12 04 58

app/components/Swapbox.tsx
berteotti: I feel like this one doesn't need abstraction since it's being used in 2 places. If we need it 1 more time lets abstract it.

ElRodrigote (approved):

app/layout.tsx
```diff
@@ -6,8 +6,12 @@ import './globals.css';
 import '@swapr/ui/colors.css';
 
 import { Providers } from '@/providers';
-import { Navbar, Footer } from '@/app/components/ui';
+
+// exporting from components directory will create an error
```
Diogomartf, line 10: why?

and why do we need this comment?

providers/providers.tsx
```diff
@@ -4,32 +4,40 @@ import { PropsWithChildren, useMemo } from 'react';
...
+  const { theme } = useTheme();
+  const connectkitTheme = (theme || 'auto') as Types.Mode;
+  const options = { enforceSupportedChains: true };
+
+  return (
+    <ConnectKitProvider mode={connectkitTheme} options={options}>
+      {children}
+    </ConnectKitProvider>
   );
 };
```
Diogomartf, line 43: why have this here?

Diogomartf (approved):

providers/providers.tsx
berteotti: So we can change connect kit theme accordingly to our theme.

app/layout.tsx
berteotti: Fixed it, missing "use client" in some components led to errors

PR #73, opened 2024-08-09
feat: disable swapbox when market has no liquidity

berteotti:
image

### DIFF
--- a/app/markets/MarketDetails.tsx
+++ b/app/markets/MarketDetails.tsx
@@ -90,7 +90,7 @@ export const MarketDetails = ({ id }: MarketDetailsProps) => {
           </div>
           {tab === 'bet' && (
             <div className="p-2">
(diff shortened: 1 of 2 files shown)


ElRodrigote (approved):

app/markets/MarketDetails.tsx
```diff
@@ -90,7 +90,7 @@ export const MarketDetails = ({ id }: MarketDetailsProps) => {
           </div>
           {tab === 'bet' && (
             <div className="p-2">
-              {!marketModel.isClosed ? (
+              {!marketModel.isClosed && marketModel.hasLiquidity ? (
```
Diogomartf, line 93: should we show a diff message like "market has no liquidity" ?

I think that might make sense

Diogomartf (approved):

PR #77, opened 2024-08-21
[fix] bug showing 0.000 on very small numbers

Diogomartf:
improve formatEtherWithFixedDecimals to fix showing 0.000 on very small numbers

Now it shows `<0.001`

image
image
image
image
image

### DIFF
--- a/app/components/CardBet.tsx
+++ b/app/components/CardBet.tsx
@@ -21,6 +21,7 @@ import { ModalId, useModal } from '@/context/ModalContext';
 import { TransactionModal } from './TransactionModal';
 import { MarketThumbnail } from './MarketThumbnail';
 import { UserPositionComplete } from '@/app/my-bets/page';
+import { formatValueWithFixedDecimals } from '@/utils';
 
 interface BetProps {
   userPositionComplete: UserPositionComplete;
@@ -47,8 +48,6 @@ export const CardBet = ({ userPositionComplete }: BetProps) => {
   const outcomeBalance = tradesOutcomeBalance({
     fpmmTrades: userPositionComplete?.fpmmTrades,
   });
-
-  const balance = outcomeBalance ? outcomeBalance.toFixed(2) : '-';
 
   if (!market) return;
 
@@ -132,7 +131,7 @@ export const CardBet = ({ userPositionComplete }: BetProps) => {
             <div className="flex items-center space-x-1">
               <p className="text-sm font-semibold text-text-med-em">Bet amount:</p>
               <p className="text-sm font-semibold text-text-high-em">
-                ${collateralAmountUSDSpent?.toFixed(2)}
+                ${formatValueWithFixedDecimals(collateralAmountUSDSpent, 4)}
               </p>
             </div>
             <div className="flex items-center space-x-1">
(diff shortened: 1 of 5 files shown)


utils/currencies.ts
```diff
@@ -9,8 +9,11 @@ export const formattedNumberDollars = (number: number) => formatter.format(numbe
 
 export const formatEtherWithFixedDecimals = (wei: bigint, fixedDecimals: number = 3) => {
   const sharesFormatted = formatEther(wei);
+  const isAnInteger = Number(sharesFormatted) % 1 === 0;
+  const isVerySmallNumber = Number(sharesFormatted) < 0.001;
```
berteotti, line 13: Here we need to take into account the `fixedDecimals` value. I suggest using an exponential where  `fixedDeciamls` (defaults to 3) is the power and 0.1 is the base. In this case, it will return 0.001.

Diogomartf: why we have to take that into account?

doesn't any number that falls bellow `0.001`  enters this category and therefore we want to show "<0.001" ?

ElRodrigote (approved):

app/components/UserBets.tsx
```diff
@@ -179,9 +180,14 @@ export const UserBets = ({ market }: UserBets) => {
...
-          const tradedBalance = outcomesTradedBalance[index].toFixed(2);
-
+          const collateralSpent = formatEtherWithFixedDecimals(
+            parseEther(outcomesCollateralAmountUSDSpent[index].toFixed(20)),
+            4
+          );
+          const tradedBalance = formatEtherWithFixedDecimals(
+            parseEther(outcomesTradedBalance[index].toFixed(20)),
+            4
+          );
```
berteotti, line 190: I think here we are complicating just to feat the function parameters. What about having two functions? One for normal numbers and other for wei?

app/components/CardBet.tsx
```diff
@@ -142,11 +146,19 @@ export const CardBet = ({ userPositionComplete }: BetProps) => {
...
+                    </p>
                     <TokenLogo address={market.data.collateralToken} className="size-3" />
                   </>
                 ) : (
-                  '$' + collateralAmountUSDSpent?.toFixed(2)
+                  '$' +
+                  formatEtherWithFixedDecimals(
+                    parseEther(collateralAmountUSDSpent?.toFixed(20)),
+                    4
+                  )
```
berteotti, line 161: Same has https://github.com/SwaprHQ/presagio/pull/77/files#r1727299519

app/components/CardBet.tsx
```diff
@@ -142,11 +146,19 @@ export const CardBet = ({ userPositionComplete }: BetProps) => {
               <div className="flex items-center space-x-1 text-sm font-semibold text-text-high-em">
                 {!market.isClosed || isWinner ? (
                   <>
-                    <p>{balance}</p>
+                    <p>
+                      {formatEtherWithFixedDecimals(
+                        parseEther(outcomeBalance.toFixed(20))
+                      )}
```
berteotti, line 152: Same has https://github.com/SwaprHQ/presagio/pull/77/files#r1727299519

app/components/CardBet.tsx
```diff
@@ -132,7 +132,11 @@ export const CardBet = ({ userPositionComplete }: BetProps) => {
             <div className="flex items-center space-x-1">
               <p className="text-sm font-semibold text-text-med-em">Bet amount:</p>
               <p className="text-sm font-semibold text-text-high-em">
-                ${collateralAmountUSDSpent?.toFixed(2)}
+                $
+                {formatEtherWithFixedDecimals(
+                  parseEther(collateralAmountUSDSpent?.toFixed(20)),
+                  4
+                )}
```
berteotti, line 139: Same has https://github.com/SwaprHQ/presagio/pull/77/files#r1727299519

app/components/UserBets.tsx
berteotti: We are making the exception as the rule. I would create two functions, one that fixes decimals, and on that formatsEther and call the on that fixes decimals.

Diogomartf: that makes sense, I'll explore that.

berteotti (approved):
Looking good!

ElRodrigote (approved):
Awesome refactors!

PR #80, opened 2024-08-29
[CHORE] - Update package dependencies

ElRodrigote:
## Description
* Updates our swapr-ui package dependency to consume the latest published version.


app/components/ConnectButton.tsx
```diff
@@ -6,15 +6,15 @@ import { useBalance, useEnsAvatar } from 'wagmi';
...
+import { Button, ButtonProps } from '@swapr/ui';
 import { PropsWithChildren } from 'react';
 import { cx } from 'class-variance-authority';
 
 interface CustomConnectButtonProps {
   address: Address;
   onClick: () => void;
   ensName?: string;
-  size?: ButtonSizeProp;
+  size?: ButtonProps['size'];
```
Diogomartf, line 17: why like this, seems a bit odd or is it the common pattern?

Diogomartf (approved):

app/components/ConnectButton.tsx
ElRodrigote: It's part of the work done in https://github.com/SwaprHQ/swapr-ui/pull/24, we had some discussions offline with @berteotti on how to handle the task, and we decided that the cleanest and simplest way to do it was to have a single source of truth being the ButtonProps interface and then consume the properties on demand.

berteotti (approved):

PR #81, opened 2024-08-30
[My bets] update filter unredeemed and related logic

Diogomartf:
- fixes a false positive
- improves names
- update Position model (add, id, condition, getOutcomeIndex)
- update Market model ( add canClaim)

### DIFF
--- a/app/my-bets/page.tsx
+++ b/app/my-bets/page.tsx
@@ -3,12 +3,11 @@
 import { CardBet, LoadingCardBet } from '@/app/components/CardBet';
 import NoBetsPage from '@/app/my-bets/NoBetsPage';
 import NoWalletConnectedPage from '@/app/my-bets/NoWalletConnectedPage';
-import { Market, Position, tradesOutcomeBalance } from '@/entities';
+import { Market, Position } from '@/entities';
 
 import { getUserPositions } from '@/queries/conditional-tokens';
-import { UserPosition } from '@/queries/conditional-tokens/types';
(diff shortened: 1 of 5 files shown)


app/my-bets/page.tsx
```diff
@@ -108,27 +106,19 @@ export default function MyBetsPage() {
...
+  const filterUnredeemedBets = useMemo(
+    () =>
+      userPositionsComplete?.filter(userPosition => {
+        const marketModel = new Market(userPosition.market);
+        const position = new Position(userPosition.position);
+        const outcomeIndex = position.getOutcomeIndex();
+        const canClaim = marketModel.canClaim(outcomeIndex, position.condition);
+        const alreadyClaimed = canClaim && +userPosition.balance === 0;
+        const canRedeem = canClaim && !alreadyClaimed;
+        return canRedeem;
```
ElRodrigote, line 118: WDYT about this:
I think this logic can be simplified, because now `canRedeem` checks if `canClaim && !(canClaim && +userPosition.balance === 0)`, but what I understand is that the user can redeem if they can claim and has balance, right?

```suggestion
        const canRedeem = canClaim && +userPosition.balance > 0;

        return canRedeem;
```

berteotti (approved):

app/my-bets/page.tsx
Diogomartf: I will keep it like this, as it will change in upcoming PR.  We'll have to support bigInt and numbers on this function.

ElRodrigote (approved):

PR #82, opened 2024-09-02
Improve redeemed related logic

Diogomartf:
- rename market to a more specific name: FixedProductMarketMaker
- fixes bug CardBet showing redeem button when already claimed
- on Market ententy rename market to fpmm
- add new entity MarketCondition to handle condition related like canRedeem, canClaim
 - remove those methods from market
- refactor UserBets to use tradesCollateralAmountUSDSpent
- extract getOutcomeUserTrades to Trade entity
- improves names
- add id to position on Omen Markets query

### DIFF
--- a/app/my-bets/page.tsx
+++ b/app/my-bets/page.tsx
@@ -109,10 +109,12 @@ export default function MyBetsPage() {
   const filterUnredeemedBets = useMemo(
     () =>
       userPositionsComplete?.filter(userPosition => {
-        const market = new Market(userPosition.fpmm);
         const position = new Position(userPosition.position);
         const outcomeIndex = position.getOutcomeIndex();
-        const marketCondition = new MarketCondition(market, position.condition);
+        const marketCondition = new MarketCondition(
+          userPosition.fpmm,
+          position.condition
+        );
         const canRedeem = marketCondition.canRedeem(outcomeIndex, userPosition.balance);
         return canRedeem;
       }) ?? [],
--- a/app/my-bets/page.tsx
+++ b/app/my-bets/page.tsx
@@ -3,7 +3,7 @@
 import { CardBet, LoadingCardBet } from '@/app/components/CardBet';
 import NoBetsPage from '@/app/my-bets/NoBetsPage';
 import NoWalletConnectedPage from '@/app/my-bets/NoWalletConnectedPage';
-import { Market, Position } from '@/entities';
+import { Market, MarketCondition, Position } from '@/entities';
 
 import { getUserPositions } from '@/queries/conditional-tokens';
 import { UserPosition, Condition } from '@/queries/conditional-tokens/types';
@@ -21,7 +21,7 @@ import { useAccount } from 'wagmi';
 
 export interface UserPositionComplete extends UserPosition {
   fpmmTrades: FpmmTrade[];
-  market: FixedProductMarketMaker;
+  fpmm: FixedProductMarketMaker;
   condition: Condition;
 }
 
@@ -60,13 +60,13 @@ export default function MyBetsPage() {
 
               const tradesData = await getMarketUserTrades({
                 creator: address.toLowerCase(),
-                fpmm: market.data.id,
+                fpmm: market.fpmm.id,
                 outcomeIndex_in: [outcomeIndex],
               });
 
               return {
                 ...userPosition,
-                market: market.data,
+                fpmm: market.fpmm,
                 condition: position.condition as Condition,
                 fpmmTrades: tradesData?.fpmmTrades || [],
               };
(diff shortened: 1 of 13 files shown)


berteotti (commented):
I see sometimes we call FixedProductMarketMaker as marketOmenFpmm, other as market, others as fpmm. It would be good to pick a name and call the same everywhere. We can even call it `fixedProductMarketMaker` as the type suggests.

Diogomartf:
> fixedProductMarketMaker



> I see sometimes we call FixedProductMarketMaker as marketOmenFpmm, other as market, others as fpmm. It would be good to pick a name and call the same everywhere. We can even call it `fixedProductMarketMaker` as the type suggests.

yeah, that makes sense. I'll check if I can uniform all of those values.

app/my-bets/page.tsx
```diff
@@ -108,27 +106,18 @@ export default function MyBetsPage() {
...
-  }, [filterCompleteBets]);
+  const filterUnredeemedBets = useMemo(
+    () =>
+      userPositionsComplete?.filter(userPosition => {
+        const market = new Market(userPosition.fpmm);
+        const position = new Position(userPosition.position);
+        const outcomeIndex = position.getOutcomeIndex();
+        const marketCondition = new MarketCondition(market, position.condition);
+        const canRedeem = marketCondition.canRedeem(outcomeIndex, userPosition.balance);
+        return canRedeem;
```
ElRodrigote, line 117: Ahhh now I get what you meant before, much more clear now hahaha

entities/markets/market.ts
```diff
@@ -2,36 +2,37 @@ import { FixedProductMarketMaker } from '@/queries/omen';
 import { fromHex } from 'viem';
 import { Outcome } from '@/entities';
 import { isPast } from 'date-fns';
+import { Condition } from '@/queries/conditional-tokens/types';
```
ElRodrigote, line 5: We can safely remove this, we're not using this here I think.

entities/markets/marketCondition.ts
```diff
@@ -0,0 +1,30 @@
...
+  alreadyClaimed(index: number, outcomeBalance: BigInt | string) {
+    const zero = typeof outcomeBalance === 'bigint' ? BigInt(0) : 0;
+    const balance = typeof outcomeBalance === 'string' ? +outcomeBalance : outcomeBalance;
+
+    return this.canClaim(index) && balance === zero;
+  }
+
+  canRedeem(index: number, outcomeBalance: any) {
+    return this.canClaim(index) && !this.alreadyClaimed(index, outcomeBalance);
+  }
```
ElRodrigote, line 29: We could merge these two functions because they do a similar process. Maybe returning a different non-boolean value:

`canClaim && balance === zero` -> claimed
`canClaim && balance !== zero` -> redeemable
`!canClaim` -> early return

queries/omen/markets.ts
berteotti, line 1: Why is this diff showing up? I think it's better to rebase you branch with `main`

app/filters.tsx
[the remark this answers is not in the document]
Diogomartf: will check

ElRodrigote (changes requested):
Needs to solve the merge conflicts

ElRodrigote (approved):

entities/markets/marketCondition.ts
Diogomartf: Having the 2 functions is handy, because we need to know if already claimed to show some things. 
We could use canRedeem but it would not be as descriptive as the name. 

for eg:

```js
 const resultString = alreadyClaimed
            ? 'You claimed:'
            : canClaim
              ? 'You won:'
              : isResolved
                ? 'You lost:'
                : 'You can win:';

```

Already claimed is very readable. I do not bother repeating a bit in this case.

entities/markets/marketCondition.ts
```diff
@@ -0,0 +1,30 @@
+import { Market } from '@/entities/markets/market';
+import { Condition as ConditionalTokenCondition } from '@/queries/conditional-tokens/types';
+
+export class MarketCondition {
```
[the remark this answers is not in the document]
Diogomartf: Not sure about it, there are some potential issues:
- it would create a tight coupling between Market and MarketCondition.
- MarketCondition would inherit all properties and methods from Market, which is not very relevant. (only isWinner)
- It could lead to confusion about whether a particular instance represents a market or a market condition.

the good part
- it removes the need to receive a market model and can just receive the `fpmm` as the MarketModel
- calling isWinner would be more simple, `this.isWinner`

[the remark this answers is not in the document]
Diogomartf: After talking with Leo, I converted MarketCondition to extend Market, as we already pass a market there there's already a relationship so we can leverage the extend and make it more powerful and simple.

berteotti (approved):

PR #83, opened 2024-09-02
feat: create better profile

berteotti:
- Adds the new profile page 
- Fixes background color too

image


utils/currencies.ts
```diff
@@ -21,7 +21,7 @@ export const formatValueWithFixedDecimals = (
   const smallNumberString = `<${smallestNumber.toFixed(fixedDecimals)}`;
   const isVerySmallNumber = Number(value) < smallestNumber;
 
-  if (isAnInteger) return value;
+  if (isAnInteger) return value.toString();
```
Diogomartf, line 24: why?

app/components/CardBet.tsx
```diff
@@ -158,22 +124,78 @@ export const CardBet = ({ userPositionComplete }: BetProps) => {
...
+
+  const outcomeIndex = position.outcomeIndex - 1;
+  const condition = userPositionComplete.position.conditions[0];
+
+  const isClaimed = +userPositionComplete.balance === 0;
+  const isResolved = condition.resolved;
+  const hasPayoutDenominator = +condition.payoutDenominator > 0;
+  const isWinner = market.isWinner(outcomeIndex);
+
+  const canClaim = isWinner && isResolved && !isClaimed && hasPayoutDenominator;
```
Diogomartf, line 155: let's use the new functions 👀

app/profile/page.tsx
```diff
@@ -0,0 +1,303 @@
...
+    if (market.answer === null) return acc;
+
+    const amountSpentInUSD = tradesCollateralAmountUSDSpent({
+      fpmmTrades: userPosition.fpmmTrades,
+    });
+
+    return amountSpentInUSD + acc;
+  }, 0);
+
+  const winsInUSD = userPositions?.reduce((acc, userPosition) => {
```
Diogomartf, line 105: is this a total?

```suggestion
  const winsTotalInUSD = userPositions?.reduce((acc, userPosition) => {
```

something more descriptive would be better

app/profile/page.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+  const positionValue =
+    potentialWinsInUSD !== undefined
+      ? formatValueWithFixedDecimals(potentialWinsInUSD, 2)
+      : '-';
+  const volumeTraded =
+    totalVolumeInUSD !== undefined
+      ? formatValueWithFixedDecimals(totalVolumeInUSD, 2)
+      : '-';
+  const succesRate =
```
Diogomartf, line 179: typo
```suggestion
  const successRate =
```

app/profile/page.tsx
```diff
@@ -0,0 +1,303 @@
...
+  );
+};
+
+const StatsCardLoading = () => (
+  <div className="w-full space-y-2 rounded-16 bg-surface-surface-0 p-6 font-semibold text-text-low-em ring-1 ring-outline-base-em">
+    <div className="h-3 w-20 animate-pulse rounded-8 bg-outline-low-em" />
+    <div className="flex space-x-2">
+      <div className="h-9 w-28 animate-pulse rounded-8 bg-outline-low-em" />
+      <div className="h-9 w-10 animate-pulse rounded-8 bg-outline-low-em" />
+    </div>
```
Diogomartf, line 301: I would remove this one

providers/duneAnalytics.ts
Diogomartf, line 1: not sure if I think this is a provider, when I think of a provider, I think something that receives a children and provides something underneath. 

But also don't have a good suggestion to where should this code go.

utils/currencies.ts
berteotti: We were returning `number | string` type, let's just return string since in the worst case this function returns `<0.01` which is a sting. So this function only returns `string`.

app/profile/page.tsx
berteotti: I like this spacing + div since it represents the symbol and looks more like the StatsCard component. One thing we can do is to just "load" the number part and show the title and symbol content. What do you think?

providers/duneAnalytics.ts
berteotti: Yeah, I also don't have a good place for this. What about `utils/duneAnalytics.ts`?

Diogomartf: that might work, I prefer that to providers!

app/profile/page.tsx
Diogomartf: not sure, would have to try it to have opinion, but go ahead and see if that feels better.

utils/currencies.ts
Diogomartf: thanks

app/components/CardBet.tsx
```diff
@@ -17,22 +17,22 @@ import {
...
 }
 
-export const CardBet = ({ userPositionComplete }: BetProps) => {
-  const [txHash, setTxHash] = useState('');
-  const [isTxLoading, setIsTxLoading] = useState(false);
-  const { openModal } = useModal();
-  const config = useConfig();
+export const CardBet = ({ userPositionComplete, children }: CardBetProps) => {
+  const position = new Position(userPositionComplete.position);
+  const outcomeIndex = position.outcomeIndex - 1;
```
Diogomartf, line 33: We should rely on position for getting `outcomeIndex` so we don't have this logic on components

```suggestion
  const outcomeIndex = position.getOutcomeIndex()
```

app/profile/page.tsx
```diff
@@ -0,0 +1,299 @@
...
+        fpmmTrades: userPosition.fpmmTrades,
+      }) + acc,
+    0
+  );
+
+  const numberOfWonBets = userPositions?.reduce((acc, userPosition) => {
+    const position = new Position(userPosition.position);
+    const market = new Market(userPosition.fpmm);
+
+    if (market.answer !== null && market.isWinner(position.outcomeIndex - 1))
```
Diogomartf, line 153:
```suggestion
    if (market.answer !== null && market.isWinner(position.getOutcomeIndex()))
```

app/profile/page.tsx
```diff
@@ -0,0 +1,299 @@
...
+      return acc + 1;
+
+    return acc;
+  }, 0);
+
+  const numberOfClosedBets = userPositions?.reduce((acc, userPosition) => {
```
Diogomartf, line 159: same, wondering if would be better to have this logic on market instead on component

app/profile/page.tsx
```diff
@@ -0,0 +1,299 @@
...
+  const volumeTraded =
+    totalVolumeInUSD !== undefined
+      ? formatValueWithFixedDecimals(totalVolumeInUSD, 2)
+      : '-';
+  const successRate =
+    numberOfClosedBets !== undefined &&
+    numberOfClosedBets !== 0 &&
+    numberOfWonBets !== undefined
+      ? ((numberOfWonBets / numberOfClosedBets) * 100).toFixed(0)
+      : '-';
```
Diogomartf, line 184: same, this probably could go to the Market entity  

probably will do the "-" here, but the logic to return the successRate.

queries/conditional-tokens/queries.ts
```diff
@@ -124,6 +125,16 @@ const getConditionQuery = gql`
   }
 `;
 
+const getUserQuery = gql`
+  query GetCondition($id: ID!) {
```
Diogomartf, line 129: this name seems wrong

```suggestion
  query GetUser($id: ID!) {
```

app/profile/page.tsx
```diff
@@ -0,0 +1,299 @@
...
+
+  const totalVolumeInUSD = userPositions?.reduce(
+    (acc, userPosition) =>
+      tradesVolume({
+        fpmmTrades: userPosition.fpmmTrades,
+      }) + acc,
+    0
+  );
+
+  const numberOfWonBets = userPositions?.reduce((acc, userPosition) => {
```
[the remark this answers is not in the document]
berteotti: Maybe we can have a `MarketTrades` class, which has **some logic** related to this. But I would do it on another PR. Relative to the entire `reduce` logic, I would leave it here for now, it's only used here so no benefit of having it on an entity. I would also create a new `UserBet` entity for that.

app/profile/page.tsx
berteotti: https://github.com/SwaprHQ/presagio/pull/83#discussion_r1745598437 answered here

app/profile/page.tsx
berteotti: https://github.com/SwaprHQ/presagio/pull/83#discussion_r1745598437 same as here

Diogomartf (approved):

PR #85, opened 2024-09-03
Add finalizing filter

kongzii:
I'm currently working on a `challenger agent`, it's supposed to challenge questions on Reality, and I was missing this filter.

In the current version, there is no way to search for a market that is answered but not yet finalized.

The `Pending` filter is great, but that will show only markets without any answer on Reality.

Current dev:

Screenshot by Dropbox Capture

Local test:

Screenshot by Dropbox Capture

Reality:

Screenshot by Dropbox Capture

### DIFF
--- a/app/filters.tsx
+++ b/app/filters.tsx
@@ -75,6 +75,7 @@ export const stateFilters: StateFilter[] = [
     when: {
       openingTimestamp_lt: nowTimestamp,
       isPendingArbitration: false,
+      currentAnswer_not: null,
     },
   },
   {
--- a/app/filters.tsx
+++ b/app/filters.tsx
@@ -70,6 +70,14 @@ export const stateFilters: StateFilter[] = [
     },
   },
   {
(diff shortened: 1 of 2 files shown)


app/filters.tsx
```diff
@@ -69,6 +69,14 @@ export const stateFilters: StateFilter[] = [
       answerFinalizedTimestamp_lt: nowTimestamp,
     },
   },
+  {
+    name: 'Finalizing',
```
kongzii, line 73: I'm not sure what the correct name for it is; it would just be nice if we could get to these markets somehow.

berteotti:
@kongzii Do you think we should have a restricter finalizing filter? Currently the filter shows pending and finalizing markets at the same time. If we also search for a `currentAnswer` different from `null` it will only show finalizing markets. What do you think about it?

kongzii:
@berteotti Fixed; thanks for noticing!

PR #87, opened 2024-09-05
feat: fetch collateral token USD price

berteotti:
Fetches token USD prices to show accurate user betting stats.

### DIFF
--- a/app/profile/page.tsx
+++ b/app/profile/page.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import { useSearchParams } from 'next/navigation';
-import { Suspense, useMemo } from 'react';
+import { useMemo } from 'react';
 import { formatEther, isAddress } from 'viem';
 import {
   calcSellAmountInCollateral,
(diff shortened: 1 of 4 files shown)


app/profile/page.tsx
```diff
@@ -52,6 +56,35 @@ export default function ProfilePage() {
...
+          queryFn: async () => await getTokenUSDPrice(collateralTokenAddress),
+          staleTime: Infinity,
+        });
+
+        setTokenPrices(tokenPrices => ({
+          [collateralTokenAddress]: price,
+          ...tokenPrices,
+        }));
+      }
+      setIsLoadingTokenPrices(false);
```
berteotti, line 84: Not sure if we should do this here or call it inside `getUserPositionsComplete` and returned it inside `UserPositionComplete`.

app/profile/page.tsx
```diff
@@ -52,6 +56,35 @@ export default function ProfilePage() {
...
+    })();
+  }, [queryClient, userPositions]);
```
berteotti, line 86: Not sure if we should do this here or call it inside `getUserPositionsComplete` and returned it inside `UserPositionComplete`.

Diogomartf (approved):

PR #89, opened 2024-09-06
Add support for ens names and avatars

Diogomartf:
This PR adds support for showing ENS and ENS avatar when those exist.

When no avatar It shows a gradient based on address

[Check out a market with ENS here](https://pgio-82-show-addresses-ens-a.presagio.pages.dev/markets?id=0xaa1c93b800a8d2b47499042af76bb9d6ff062cd7)

image

- add new UserAvatarAddress component
- rename `ChainId.ETHEREUM` to `ChainId.MAINNET` to be consistent with wagmi lib

With avatars
image

Added to profile page
image


app/components/UserAvatarAddress.tsx
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }, [address]);
+
+  const { data: ensName } = useEnsName({
+    address: normalizedAddress,
+    chainId: mainnet.id,
+  });
+
+  const avatarGradientUrl = (address: string) =>
+    `https://avatars.jakerunzer.com/${address}`;
```
berteotti, line 33: What is this URL? Maybe we should give it a name

providers/chain-config.ts
```diff
@@ -1,19 +1,20 @@
 import { createConfig, fallback, http } from 'wagmi';
 import { getDefaultConfig } from 'connectkit';
-import { gnosis } from 'wagmi/chains';
+import { gnosis, mainnet } from 'wagmi/chains';
 import { safe } from 'wagmi/connectors';
 import { ChainId, RPC_LIST } from '@/constants';
 
 const defaultConfig = getDefaultConfig({
-  chains: [gnosis],
+  chains: [gnosis, mainnet],
```
berteotti, line 8: Do we need this? I'm not sure if we need it. This makes it possible to connect to ethereum and not show some kind of error. Example screenshot:

image

berteotti (commented):
Left a few comments, design looking good.

providers/chain-config.ts
Diogomartf: the only way I found to make the ens work, cause they have to pull data from `mainnet` and not having it on config would only make it work on your own ens.

app/components/UserAvatarAddress.tsx
Diogomartf: it's the one that give us the gradient based on the address, I'll name it.

Diogomartf: I'll go with `GRADIENT_BASED_ON_ADDRESS_URL`

providers/chain-config.ts
berteotti: Took a quick look, let's remove mainnet from this default config.

And let's create a new config, called `mainnetConfig` and use it with `useEnsName` only.

Here is the example:

```
export const mainnetConfig = createConfig({
  chains: [mainnet],
  transports: {
    [mainnet.id]: fallback([http(RPC_LIST[ChainId.MAINNET]), http()]),
  },
});
```

Then let's pass it on `useEnsName`

```
const { data: ensName } = useEnsName({
    address: normalizedAddress,
    chainId: mainnet.id,
    config: mainnetConfig,
  });
  ```

File: app/components/UserAvatarAddress.tsx [hunk 3]
```diff
@@ -0,0 +1,82 @@
...
+    config: mainnetConfigForENS,
+  });
+
+  const { data: ensAvatar } = useEnsAvatar({
+    name: ensName ?? undefined,
+    chainId: mainnet.id,
+    config: mainnetConfigForENS,
+  });
+
+  console.log('ensAvatar:', ensAvatar);
```
berteotti, line 45: console.log

app/components/UserAvatarAddress.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const GRADIENT_BASED_ON_ADDRESS_URL = 'https://avatars.jakerunzer.com';
+  const ENS_METADATA_AVATAR_URL = 'https://metadata.ens.domains/mainnet/avatar';
+
+  const avatarGradientUrl = (address: string) =>
+    `${GRADIENT_BASED_ON_ADDRESS_URL}/${address}`;
+
+  const avatarEnsUrl = (ens: string) => `${ENS_METADATA_AVATAR_URL}/${ens}`;
+
+  const avatarUrl =
+    ensName && ensAvatar ? avatarEnsUrl(ensName) : avatarGradientUrl(normalizedAddress);
```
berteotti, line 56: Should we have a loading for ensName and ensAvatar?

Diogomartf: I think it's okay since they already have a default  "bg gray", so it sense that isLoading

app/components/AddressLink.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { shortenAddress } from '@/utils';
+import Image from 'next/image';
+import Link from 'next/link';
+import { useMemo } from 'react';
+import { twMerge } from 'tailwind-merge';
+import { Address, getAddress, http } from 'viem';
+import { mainnet } from 'viem/chains';
+import { useEnsName } from 'wagmi';
+
+interface AddressLinkProps {
```
berteotti, line 13: Let `AddressLinkProps` extend from `Link` props, and spread the rest of the props to the link component. This to allow passing link props to the component.

app/components/AddressLink.tsx
```diff
@@ -0,0 +1,58 @@
...
+      console.error('Invalid Ethereum address:', error);
+      return address;
+    }
+  }, [address]);
+
+  const { data: ensName } = useEnsName({
+    address: normalizedAddress,
+    chainId: mainnet.id,
+    config: mainnetConfigForENS,
+  });
```
berteotti, line 41: I'm almost convinced that this should be a hook that we can use wherever we need.

app/profile/page.tsx
```diff
@@ -213,18 +212,14 @@ export default function ProfilePage() {
...
-            rel="noopener noreferrer"
-          >
-            {shortenAddress(address)}
-          </a>
-          {isAIAgent && <Image src="/ai.svg" alt="ai" width={26} height={26} />}
+            address={address}
+            isAIAgent={isAIAgent}
+            className="text-2xl font-semibold"
+            iconSize={24}
+          />
```
berteotti, line 222: Missing `target="_blank"`

app/components/AddressLink.tsx
Diogomartf: I'll explore this idea and see how it feels.

berteotti (approved):

PR #90, opened 2024-09-10
feat: show if current answer is not resolved

berteotti:
image

### DIFF
--- a/app/filters.tsx
+++ b/app/filters.tsx
@@ -4,6 +4,7 @@ import {
   OrderDirection,
 } from '@/queries/omen';
 import { AI_AGENTS_ALLOWLIST } from '../constants';
+import { _24HoursInSeconds, nowTimestamp } from '@/utils/time';
 
 export type OrderFilter = {
   name: string;
@@ -45,13 +46,13 @@ export const orderFilters: OrderFilter[] = [
   },
 ];
 
-const nowTimestamp = Math.floor(Date.now() / 1000);
-
 export type StateFilter = {
   name: string;
   key: string;
   when: FixedProductMarketMaker_Filter;
 };
+
+const oneDayAgoTimestamp = nowTimestamp - _24HoursInSeconds;
 
 export const stateFilters: StateFilter[] = [
   {
@@ -73,6 +74,8 @@ export const stateFilters: StateFilter[] = [
     name: 'Finalizing',
     key: 'finalizing',
     when: {
+      resolutionTimestamp: null,
+      currentAnswerTimestamp_gt: oneDayAgoTimestamp,
       openingTimestamp_lt: nowTimestamp,
       isPendingArbitration: false,
       currentAnswer_not: null,
--- a/app/markets/Bet.tsx
+++ b/app/markets/Bet.tsx
@@ -18,7 +18,7 @@ export const Bet = ({ fixedProductMarketMaker }: BetProps) => {
 
   if (
     !fixedProductMarketMaker.isPendingArbitration &&
-    marketModel.answer === null &&
+    marketModel.currentAnswer === null &&
     marketModel.isClosed
   )
     return (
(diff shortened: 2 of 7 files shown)


app/filters.tsx
```diff
@@ -73,6 +72,8 @@ export const stateFilters: StateFilter[] = [
     name: 'Finalizing',
     key: 'finalizing',
     when: {
+      resolutionTimestamp: null,
+      currentAnswerTimestamp_gt: nowTimestamp - _24HoursInSeconds,
```
Diogomartf, line 76: I think we could encapsulate this on utils, `oneDayAgoTimestamp` something like this

entities/markets/market.ts
```diff
@@ -2,20 +2,31 @@ import { FixedProductMarketMaker } from '@/queries/omen';
 import { fromHex } from 'viem';
 import { Outcome } from '@/entities';
 import { isPast } from 'date-fns';
+import { _24HoursInSeconds, nowTimestamp } from '@/utils/time';
```
Diogomartf, line 5:
```suggestion
import { _24HoursInSeconds, nowTimestamp } from '@/utils';
```

Diogomartf:
is there a market we can test?

The telegram on the screenshot seems to be resolved already.

berteotti:
> is there a market we can test?

You can use the finalizing filter to check those finalizing markets.
image

Also, here is an example market https://73d36dd6.presagio.pages.dev/markets?id=0x8436c565563d7dbfe8f1696d295b57ec16112250

app/markets/Bet.tsx
```diff
@@ -60,6 +60,36 @@ export const Bet = ({ fixedProductMarketMaker }: BetProps) => {
...
+        <div className="flex items-center space-x-2">
+          <span>Current winner outcome is</span>
+          <Tag
+            className="w-fit uppercase"
+            colorScheme={marketModel.currentAnswer === 0 ? 'success' : 'danger'}
+          >
+            {marketModel.outcomes[marketModel.currentAnswer]?.name}
+          </Tag>
+        </div>
+        <div className="flex flex-col items-center space-y-2">
```
Diogomartf, line 75: this will improve the styles
```suggestion
        <div className="bg-surface-surface-1 flex flex-col items-center py-4 rounded-12 w-full">
```

entities/markets/market.ts
berteotti: I had what you suggest, but had to change since it was giving me errors. I suggest you to try to actually see what's going on. This is what I don't like about NextJS :)

app/markets/Bet.tsx
berteotti: Will give it a try

app/filters.tsx
berteotti: Will create a const here in this file, if it's used elsewhere we should move it to utils

entities/markets/market.ts
Diogomartf: what kind of errors? 

usually it works like this.

app/markets/Bet.tsx
```diff
@@ -60,6 +60,36 @@ export const Bet = ({ fixedProductMarketMaker }: BetProps) => {
       </div>
     );
 
+  if (marketModel.currentAnswer !== null && !marketModel.isAnswerResolved)
```
Diogomartf, line 63: Can we get a name for this condition? I think that would be better.

Diogomartf (approved):

entities/markets/market.ts
berteotti: Strange ones, like "ReferenceError: Cannot access 'c' before initialization". I suggest you to try it.

app/markets/Bet.tsx
Diogomartf: ok, we can leave as it is for now, but wondering if it was an variable was a bit more clear.

PR #91, opened 2024-09-12
feat: add bet potential profit

berteotti:
image

### DIFF
--- a/app/components/Swapbox.tsx
+++ b/app/components/Swapbox.tsx
@@ -28,7 +28,9 @@ import {
   useReadToken,
 } from '@/hooks/contracts/erc20';
 import { gnosis } from 'viem/chains';
-import { formatEtherWithFixedDecimals } from '@/utils';
+import { formatEtherWithFixedDecimals, formatValueWithFixedDecimals } from '@/utils';
+import { useQuery } from '@tanstack/react-query';
+import { getTokenUSDPrice } from '@/queries/mobula';
 
 export const SLIPPAGE = 0.01;
 const ONE_UNIT = '1';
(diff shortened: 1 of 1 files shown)


app/components/Swapbox.tsx
```diff
@@ -230,6 +238,28 @@ export const Swapbox = ({ fixedProductMarketMaker }: SwapboxProps) => {
...
+    ? swapDirection === SwapDirection.BUY
+      ? formatValueWithFixedDecimals(collateralTokenUSDPrice, 2)
+      : formatValueWithFixedDecimals(
+          collateralTokenUSDPrice * +currentState.tokenPrice,
+          2
+        )
+    : null;
+
+  const potentialProfit =
+    tokenAmountOut && tokenAmountIn && swapDirection === SwapDirection.BUY
```
Diogomartf, line 251: `const swapDirection = SwapDirection.BUY`

```suggestion
    tokenAmountOut && tokenAmountIn && isBuying
```

Diogomartf (approved):

app/components/Swapbox.tsx
```diff
@@ -230,6 +240,28 @@ export const Swapbox = ({ fixedProductMarketMaker }: SwapboxProps) => {
...
     !supportedChains.some(supportedChain => supportedChain.id === connectorChainId);
 
+  const priceInUSD = collateralTokenUSDPrice
+    ? isBuying
+      ? formatValueWithFixedDecimals(collateralTokenUSDPrice, 2)
+      : formatValueWithFixedDecimals(
+          collateralTokenUSDPrice * +currentState.tokenPrice,
+          2
+        )
+    : null;
```
ElRodrigote, line 250: Minor, non-blocking:
Maybe we can simplify a bit the ternary operator nesting, WDYT? 
```suggestion
  const priceInUSD = collateralTokenUSDPrice
    ? formatValueWithFixedDecimals(
        collateralTokenUSDPrice * (isBuying ? 1 : +currentState.tokenPrice),
        2
      )
    : null;
```

ElRodrigote (approved):

app/components/Swapbox.tsx
berteotti: I think this is harder to read. We are mixing code with math so you don't have a easy time understanding what the calculation is. But will create a `formattedPriceInUSD` variable to simplify reading.

PR #94, opened 2024-09-13
feat: swap default market creator order

berteotti:
image

### DIFF
--- a/app/filters.tsx
+++ b/app/filters.tsx
@@ -109,13 +109,13 @@ export type CreatorFilter = {
 
 export const creatorFilters: CreatorFilter[] = [
   {
-    name: 'AI markets',
-    key: 'ai',
-    when: { creator_in: AI_AGENTS_ALLOWLIST },
-  },
-  {
     name: 'All markets',
     key: 'all',
     when: {},
+  },
+  {
+    name: 'AI markets',
+    key: 'ai',
+    when: { creator_in: AI_AGENTS_ALLOWLIST },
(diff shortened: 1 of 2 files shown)


app/filters.tsx
```diff
@@ -108,14 +108,14 @@ export type CreatorFilter = {
...
-    key: 'ai',
-    when: { creator_in: AI_AGENTS_ALLOWLIST },
-  },
   {
     name: 'All markets',
     key: 'all',
     when: {},
   },
+  {
+    name: 'AI markets',
```
Diogomartf, line 117: We don't need AI anymore here as it will have the svg

```suggestion
    name: 'markets',
```

ElRodrigote (approved):

app/page.tsx
```diff
@@ -277,7 +278,12 @@ export default function HomePage() {
                     })}
                     name="tick-fill"
                   />
-                  <p>{option.name}</p>
+                  <div className="flex items-center space-x-2">
+                    <p>{option.name}</p>
+                    {option.key === 'ai' && (
+                      <Image src="/ai.svg" alt="ai" width={20} height={20} />
```
Diogomartf, line 284: I would probably move the icon to the left instead of right.

app/filters.tsx
berteotti: Tested it and didn't work as well as expected. Will add the AI logo as it is, and we can check this again in the future.

PR #96, opened 2024-09-17
[PGIO-65][FEAT] - Create Skeleton component that abstracts styles

ElRodrigote:
## Description
* Abstracts common "loading" styles and structure into a Skeleton component
* Replace the repeated styles with the new Skeleton
* Minor helper components renaming for consistency's sake

### DIFF
--- a/app/markets/MarketActivity.tsx
+++ b/app/markets/MarketActivity.tsx
@@ -92,7 +92,7 @@ export const MarketActivity = ({ id }: { id: string }) => {
   return (
     <div>
       <div>
-        {true ? (
+        {isLoading ? (
           <LoadingMarketActivity />
         ) : (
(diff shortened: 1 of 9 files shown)


berteotti (commented):
Looking good, I think it's just missing on `TokenLogo`

berteotti (approved):

app/markets/MarketActivity.tsx
```diff
@@ -92,8 +92,8 @@ export const MarketActivity = ({ id }: { id: string }) => {
   return (
     <div>
       <div>
-        {isLoading ? (
-          <LoadingSkeleton />
+        {true ? (
```
berteotti, line 95: Should be isLoading here

ElRodrigote:
> Looking good, I think it's just missing on `TokenLogo`

As we spoke offline, applying this change means extending the `twMerge` function or abstracting the Skeleton component in a package with that extended function. We agreed not to abstract the TokenLogo loading with the Skeleton.

berteotti (changes requested):

PR #98, opened 2024-09-19
fix: pass outcomeIndex to redeem in userBets

### LINKED ISSUE #97: Hardcoded outcomeIndex in UserBets redeem
https://github.com/SwaprHQ/presagio/blob/71b76b0d4fea7cfcb90ebe7e47a00845f205675a/app/components/UserBets.tsx#L134

causes this tx to fail sometimes, when called from the /markets page. compare to this one, the one that renders on /my-bets page, which works:
https://github.com/SwaprHQ/presagio/blob/71b76b0d4fea7cfcb90ebe7e47a00845f205675a/app/components/CardBet.tsx#L155

greenlucid:
fixes #97 

I'm not actually sure if this is the approach, I didn't know how to get the position, and I don't know if the indexes of the `marketModel` are the same as the `outcomeIndex`, feel free to ignore. wasn't able to test since I'm lacking API keys to fetch subgraph / dune info.

https://github.com/greenlucid/presagio/blob/e67fb2c69c3113b74f82e706b1d242d3ff571a94/app/components/UserBets.tsx#L57
leads me to think they might be different, since it's filtering for only those outcomeIndexes for some reason? unless, that filtering is also a bug. not too familiar with this codebase tbh

### DIFF
--- a/app/components/UserBets.tsx
+++ b/app/components/UserBets.tsx
@@ -125,13 +125,13 @@ export const UserBets = ({ fixedProductMarketMaker }: UserBets) => {
   const { condition } = conditionData;
   const isResolved = condition.resolved;
 
-  const redeem = async () => {
+  const redeem = async (outcomeIndex: number) => {
     setIsTxLoading(true);
 
     try {
       const txHash = await redeemPositions({
         conditionId: condition.id,
-        outcomeIndex: 1,
+        outcomeIndex: outcomeIndex,
       });
       setTxHash(txHash);
       openModal(ModalId.WAITING_TRANSACTION);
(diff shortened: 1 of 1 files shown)


app/components/UserBets.tsx
```diff
@@ -243,7 +243,7 @@ export const UserBets = ({ fixedProductMarketMaker }: UserBets) => {
                         width="full"
                         size="lg"
                         className="space-x-2"
-                        onClick={redeem}
+                        onClick={redeem(index)}
```
Diogomartf, line 246: we need to change this to 
`() => redeem(index)`

Otherwise it will trigger the function when rendering the button vs when clicked.

Diogomartf:
There is some weird logic related to `outcomeIndex`, for the smart contract is expected to be 1 and 2, but from the subgraph they come as 0 and 1 (they probably should be normalised).

This been the biggest source of misunderstanding our codebase, but the best guess is they don't use 0 on the smart contract cause it represents a position that doesn't correspond to any outcome, also 0 could lead to division by zero errors or other unexpected behaviours in calculations and probably would need more conditions.

Diogomartf:
Thanks a lot for the report and PR, will fix the bug and simplify our approach.

You can check it here: https://github.com/SwaprHQ/presagio/pull/103

PR #99, opened 2024-09-20
feat: add highlighted market on home page

berteotti:
image


Diogomartf:
> [!TIP]
> Let's improve the "dots", make it more smooth (smaller, better colors)
> Also would be great if we could add some smooth arrows on desktop, to make it easier to swipe. 

image

Diogomartf:
We should only show markets that are open, if none of the highlighted are open, we should not show highlights.

I think we should only highlight markets that have more that 7 days remaining.

app/page.tsx
```diff
@@ -425,3 +435,71 @@ export default function HomePage() {
     </div>
   );
 }
+
+const MarketHighlight = () => {
+  const marketIdList = [
+    '0x8943a769cb0503f7bc4e92a777fcd1aa9decfc33',
+    '0x0539590c0cf0d929e3f40b290fda04b9b4a8cf68',
+  ];
```
Diogomartf, line 443: we should have a place for this outside, like a file or something, so we to update it we will only need update that file, making updates on highlights very clear.

app/page.tsx
```diff
@@ -425,3 +435,71 @@ export default function HomePage() {
     </div>
   );
 }
+
+const MarketHighlight = () => {
+  const marketIdList = [
```
Diogomartf, line 440:
```suggestion
  const highlightedMarketIdList = [
```

app/page.tsx
```diff
@@ -425,3 +435,71 @@ export default function HomePage() {
...
+                  {/* <p className="text-[44px] font-semibold leading-[56px]"> */}
+                  <p className="text-xl font-semibold">{market.title}</p>
+                </div>
+                <OutcomeBar market={market} />
+                <p className="text-base font-semibold text-text-med-em">
+                  ${formatValueWithFixedDecimals(market.usdVolume, 2)} Vol
+                </p>
+              </div>
+              <Image
+                className="h-full rounded-e-20"
```
ElRodrigote, line 495: @see 
https://nextjs.org/docs/pages/building-your-application/optimizing/images#priority

```suggestion
                priority
                className="h-full rounded-e-20"
```

Diogomartf:
we should have a default image in case there in no highlighted image or the market doesn't have one, so we can display a pattern or something for now.

I used "Lisbon" svg from https://heropatterns.com/ for an example but Akash can do a better one.

image

app/components/Carousel.tsx
```diff
@@ -0,0 +1,224 @@
+'use client';
+
+import * as React from 'react';
```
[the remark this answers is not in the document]
berteotti: We do cause we are using `React` below

app/page.tsx
berteotti: Changed to a div with a background which handles mage fallback better than `Image` component

app/components/Carousel.tsx
```diff
@@ -0,0 +1,278 @@
...
+};
+
+type CarouselContextProps = {
+  carouselRef: ReturnType<typeof useEmblaCarousel>[0];
+  api: ReturnType<typeof useEmblaCarousel>[1];
+  scrollPrev: () => void;
+  scrollNext: () => void;
+  canScrollPrev: boolean;
+  canScrollNext: boolean;
+} & CarouselProps;
```
Diogomartf, line 27: doesn't the lib provide types ?

app/page.tsx
```diff
@@ -525,3 +544,83 @@ const LoadingMarketCategories = () => (
...
+          className="h-44 w-full rounded-e-0 rounded-t-20 md:h-full md:w-1/2 md:rounded-e-20 md:rounded-s-0 2xl:w-2/5"
+          src={image}
+          alt="Market highlight"
+          style={{ objectFit: 'cover', objectPosition: 'top' }}
+          onError={() => setImage(defaultHighlightImage)}
+        />
+      </Link>
+    </CarouselItem>
+  );
+};
```
Diogomartf, line 626: I would extract these component to it's own file

market-highlight.config.ts
```diff
@@ -0,0 +1,18 @@
...
+// Image should be stored inside public/assets/highlights
+// import kamalaTrump from '@/public/assets/highlights/kamala-trump.png';
+//
+// ['0x0539590c0cf0d929e3f40b290fda04b9b4a8cf68']: { image: kamalaTrump }
+
+import kamalaTrump from '@/public/assets/highlights/kamala-trump.png';
+import { StaticImageData } from 'next/image';
+
+const highlightedMarketsList: Record<string, { image: StaticImageData }> = {
+  ['0x8943a769cb0503f7bc4e92a777fcd1aa9decfc33']: {
```
Diogomartf, line 12: why the array here?

package-lock.json
Diogomartf, line 1: remove package-lock

market-highlight.config.ts
berteotti: Removed

app/page.tsx
berteotti: Changed to use Image, applied `prioirty` thanks

market-highlight.config.ts
```diff
@@ -0,0 +1,18 @@
+// Example market highlight definition
+//
+// Image should be stored inside public/assets/highlights
+// import kamalaTrump from '@/public/assets/highlights/kamala-trump.png';
+//
+// ['0x0539590c0cf0d929e3f40b290fda04b9b4a8cf68']: { image: kamalaTrump }
```
Diogomartf, line 6:
```suggestion
// '0x0539590c0cf0d929e3f40b290fda04b9b4a8cf68': { image: kamalaTrump }
```

app/MarketHighlight.tsx
```diff
@@ -0,0 +1,101 @@
...
+  CarouselPrevious,
+  CarouselSelector,
+} from './components/Carousel';
+import Autoplay from 'embla-carousel-autoplay';
+import { highlightedMarketsList } from '@/market-highlight.config';
+import defaultHighlightImage from '@/public/assets/highlights/default.png';
+import Image from 'next/image';
+import { useState } from 'react';
+
+export const MarketHighlight = () => {
```
Diogomartf, line 22: The name should be plural as we are highlighting markets
```suggestion
export const MarketsHighlight = () => {
```

Diogomartf (approved):

PR #102, opened 2024-09-23
[PGIO-85][FEAT] - dynamically fetch categories from Dune

ElRodrigote:
## Description
* Creates a way to query Dune dashboards to fetch our market categories sorted by "popularity", meaning, the number of markets in each category.

## Visual Evidence
Screenshot 2024-09-23 at 15 10 02

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -46,7 +46,6 @@ type CategoryOptions = Categories | '';
 type DuneCategory = { Category: string };
 
 /**
- *
  * @param {DunateCategory[]} data - Response from Dune API query
  * @returns {{ [key in Categories]: number }[]} A sorted array of objects with the count of markets per category
  */
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -45,10 +45,6 @@ const SEARCH_DEBOUNCE_DELAY = 600;
 type CategoryOptions = Categories | '';
 type DuneCategory = { Category: string };
 
-/**
- * @param {DunateCategory[]} data - Response from Dune API query
- * @returns {{ [key in Categories]: number }[]} A sorted array of objects with the count of markets per category
- */
 const rankCategoriesByMarketCount = (data: any) => {
   if (!data) return [];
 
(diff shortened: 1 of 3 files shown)


app/page.tsx
```diff
@@ -29,26 +30,52 @@ import {
...
+      ({ category: categoryOption }: { category: string }) => categoryOption === category
+    );
+    const count = market ? market.cnt : 0;
+    allMarketsSortedCategories.push({ [category]: count });
+  }
+
+  allMarketsSortedCategories.sort((a, b) => Object.values(b)[0] - Object.values(a)[0]);
+
+  return allMarketsSortedCategories;
+};
```
Diogomartf, line 77: It would be great if we could simplify these code.

Diogomartf:
I think it's a good idea, what's preventing us from shipping?

Let's go with the middle one (only open markets).

Design wise, let's remove the fire emoji as it implies not a category but a trendy/HOT markets.

app/page.tsx
Diogomartf, line 77: it would be great if we could improve this 2 functions a bit

- don't think the name are good 
- probably we can simplify a bit of the code, it's not that easy to read.

app/page.tsx
```diff
@@ -29,26 +30,52 @@ import {
...
-  POLITICS = 'politics',
-  ECONOMY = 'economy',
-  INTERNATIONAL = 'international',
-  SPORTS = 'sports',
-  ENTERTAINMENT = 'entertainment',
-}
-
 type CategoryOptions = Categories | '';
 
+const groupOpenMarketsByCategory = (data: any) => {
```
Diogomartf, line 44: I think this makes a better name: `rankCategoriesByMarketCount`, it emphasizes that the categories are being ranked (sorted) based on the number of markets.

```suggestion
const rankCategoriesByMarketCount = (data: any) => {
```

app/page.tsx
```diff
@@ -31,26 +31,41 @@ import {
...
+const SEARCH_DEBOUNCE_DELAY = 600;
 
 type CategoryOptions = Categories | '';
+type DuneCategory = { Category: string };
+
+/**
+ * @param {DunateCategory[]} data - Response from Dune API query
+ * @returns {{ [key in Categories]: number }[]} A sorted array of objects with the count of markets per category
+ */
+const rankCategoriesByMarketCount = (data: any) => {
```
Diogomartf, line 52: missing the type here ?

app/page.tsx
```diff
@@ -31,26 +31,41 @@ import {
...
+const ITEMS_PER_PAGE = 12;
+const SEARCH_DEBOUNCE_DELAY = 600;
 
 type CategoryOptions = Categories | '';
+type DuneCategory = { Category: string };
+
+/**
+ * @param {DunateCategory[]} data - Response from Dune API query
+ * @returns {{ [key in Categories]: number }[]} A sorted array of objects with the count of markets per category
+ */
```
Diogomartf, line 51: I would remove the comment, as it does not add much that we can't see on the function bellow.

app/page.tsx
ElRodrigote: As spoken offline, we decided to keep the `any` type as we would need a type-guard function to check if the `unknown` in `Record<string, unknown>[]` (the data rows) value returned by the dune API, is a `Record<string, string>[]`

Diogomartf (approved):

PR #105, opened 2024-09-30
Fix: On redeem was triggering TransactionModal Success straight way without completing and loading

Diogomartf:
**Description**
New PR this fixes the problems we were having when clicking "redeem" on "My bets list" and "Market" views.
The problem was the loading modal was not showing and the success modal was triggered on click.

Now it should show proper loading tx modal and then on success it shows the green tick.
One of the problems was that we were using TxProvider which adds a `TransactionModal` component to the tree and we added another ones on `CardBet` and `UserBets` components, so when triggering openModal, there were 2 modals opening. Also some conditions on the modal needed extra checks.

This PR fixes those issues by making TxProvider more flexible by having 2 functions:

- `submitTx` (which receives the args and calls a write contract function from `wagmi`)
- `submitCustomTx` (which receives a customFunction and args that are called instead of the `writeContract` function)

In the end both do write contract calls but one can receive the function already prepared, as is the case of redeemPositions

**Summary**
add submitCustomTx to TxContext
- able to receive custom write function as an argument
- update components to use new function
- remove extra TransactionModals from components that was causing bugs

### DIFF
--- a/app/components/CardBet.tsx
+++ b/app/components/CardBet.tsx
@@ -141,12 +141,11 @@ export const MyBetsCardBet = ({ userBets }: { userBets: UserBets }) => {
   const canRedeem = marketCondition.canRedeem(outcomeIndex, userBets.balance);
 
   const redeem = async () => {
-    await submitCustomTx({
-      writeTxFunction: redeemPositions,
-      args: {
+    await submitCustomTx(() =>
+      redeemPositions({
         conditionId: condition.id,
-      },
-    });
+      })
+    );
   };
 
   return (
--- a/app/components/CardBet.tsx
+++ b/app/components/CardBet.tsx
@@ -1,8 +1,8 @@
 'use client';
 
+import { PropsWithChildren } from 'react';
 import { cx } from 'class-variance-authority';
 import Link from 'next/link';
-import { useConfig } from 'wagmi';
 
 import { Button, Tag } from '@swapr/ui';
 import { Card, TokenLogo } from '@/app/components';
@@ -17,13 +17,10 @@ import {
   UserBets,
 } from '@/entities';
 import { redeemPositions } from '@/hooks/contracts';
-import { waitForTransactionReceipt } from 'wagmi/actions';
-import { PropsWithChildren, useState } from 'react';
-import { ModalId, useModal } from '@/context/ModalContext';
 import { MarketThumbnail } from './MarketThumbnail';
 import { Skeleton } from './Skeleton';
-import { TransactionModal } from './TransactionModal';
 import { formatValueWithFixedDecimals } from '@/utils';
+import { useTx } from '@/context';
 
 interface CardBetProps extends PropsWithChildren {
   userBets: UserBets;
@@ -133,10 +130,7 @@ export const CardBet = ({ userBets, children }: CardBetProps) => {
 };
 
 export const MyBetsCardBet = ({ userBets }: { userBets: UserBets }) => {
-  const config = useConfig();
-  const [txHash, setTxHash] = useState('');
-  const [isTxLoading, setIsTxLoading] = useState(false);
-  const { openModal } = useModal();
+  const { submitCustomTx } = useTx();
 
   const position = new Position(userBets.position);
   const outcomeIndex = position.getOutcomeIndex();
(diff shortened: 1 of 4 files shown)


ElRodrigote (approved):

app/components/CardBet.tsx
```diff
@@ -147,34 +141,20 @@ export const MyBetsCardBet = ({ userBets }: { userBets: UserBets }) => {
...
 
   const redeem = async () => {
-    setIsTxLoading(true);
-
-    try {
-      const txHash = await redeemPositions({
+    await submitCustomTx({
+      writeTxFunction: redeemPositions,
+      args: {
         conditionId: condition.id,
```
berteotti, line 147:
```suggestion
await submitCustomTx(() =>
      redeemPositions({
        conditionId: condition.id,
      })
    );
```

app/components/UserBets.tsx
```diff
@@ -126,23 +120,12 @@ export const UserBets = ({ fixedProductMarketMaker }: UserBets) => {
...
 
   const redeem = async () => {
-    setIsTxLoading(true);
-
-    try {
-      const txHash = await redeemPositions({
+    await submitCustomTx({
+      writeTxFunction: redeemPositions,
+      args: {
         conditionId: condition.id,
```
berteotti, line 126:
```suggestion
await submitCustomTx(() =>
      redeemPositions({
        conditionId: condition.id,
      })
    );
```

context/TxContext.tsx
```diff
@@ -8,13 +8,19 @@ import { type WriteContractParameters } from '@wagmi/core';
...
 import { config } from '@/providers/chain-config';
+import { Address } from 'viem';
 
 export interface TxContextProps {
-  submitTx: (args: WriteContractParameters) => Promise<void>;
+  submitTx: (args: WriteContractParameters) => Promise<Address | void>;
+  submitCustomTx: (args: {
+    writeTxFunction: any;
+    args: any;
+  }) => Promise<Address | void>;
```
berteotti, line 18:
```suggestion
  submitTx: (args: WriteContractParameters) => Promise<void>;
  submitCustomTx: (tx: Function) => Promise<void>;
```

context/TxContext.tsx
```diff
@@ -8,13 +8,19 @@ import { type WriteContractParameters } from '@wagmi/core';
...
 }
 
 export const TxContext = createContext<TxContextProps>({
-  submitTx: async (args: WriteContractParameters) => {},
+  submitTx: async (args: WriteContractParameters) => '0x' as Address | void,
+  submitCustomTx: async (args: any) => '0x' as Address | void,
```
berteotti, line 23:
```suggestion
  submitTx: async () => {},
  submitCustomTx: async () => {},
```

context/TxContext.tsx
```diff
@@ -24,15 +30,22 @@ export const TxProvider = ({ children }: PropsWithChildren) => {
 
   const { openModal } = useModal();
 
-  const submitTx = async (args: WriteContractParameters): Promise<void> => {
+  const submitCustomTx = async ({
+    writeTxFunction,
+    args,
+  }: any): Promise<Address | void> => submitWriteTx(() => writeTxFunction(args));
```
berteotti, line 36:
```suggestion
  const submitCustomTx = async (tx: Function): Promise<void> => submitWriteTx(() => tx());
```

context/TxContext.tsx
```diff
@@ -24,15 +30,22 @@ export const TxProvider = ({ children }: PropsWithChildren) => {
...
+
+  const submitTx = async (args: WriteContractParameters): Promise<Address | void> =>
```
berteotti, line 38:
```suggestion
  const submitTx = async (args: WriteContractParameters): Promise<void> =>
```

context/TxContext.tsx
```diff
@@ -24,15 +30,22 @@ export const TxProvider = ({ children }: PropsWithChildren) => {
...
+    submitWriteTx(() => writeContract(config, args));
+
+  const submitWriteTx = async (writeTxFn?: any): Promise<Address | void> => {
```
berteotti, line 41:
```suggestion
  const submitWriteTx = async (writeTxFn: Function): Promise<void> => {
```

berteotti (commented):
Found a few improvements:

- Easier way to call submitCustomTx;
- Improve types.

context/TxContext.tsx
Diogomartf: Here the function name should be `writeTxFn`instead of only `tx`

Diogomartf:
good improvements, thanks @berteotti!

berteotti (approved):

Issue #101, opened 2024-09-23
Finalizing filter is using hardcoded `oneDayAgoTimestamp`


kongzii:
Hey, I noticed that the finalizing filter is using a hardcoded `oneDayAgoTimestamp` timestamp:

```
when: {
      resolutionTimestamp: null,
      currentAnswerTimestamp_gt: oneDayAgoTimestamp,
      openingTimestamp_lt: nowTimestamp,
      isPendingArbitration: false,
      currentAnswer_not: null,
    },
```

I'm afraid this works currently because all Olas and our market creators are using 24-hour finalizing periods, but this is configurable per market.

For example, if we create a very-high liquidity market, we could set the finalizing period to 1 week instead of a day to make sure there is enough time for any potential challenges.

I think this filter should work generally with 

```
    when: {
      resolutionTimestamp: null,
      answerFinalizedTimestamp_gt: nowTimestamp,
      answerFinalizedTimestamp_not: null,
      openingTimestamp_lt: nowTimestamp,
      isPendingArbitration: false,
      currentAnswer_not: null,
    },
```

But I'm not sure what bug I caused with the original filter configuration (but thanks for fixing!) and I don't have any bugged markets as an example because all of them are finalized in 24 hours at the moment. So I'm just creating an issue instead of PR.

PR #110, opened 2024-10-04
Don't hardcode 24h finalization

kongzii:
Yesterday, I created this market for some debugging: https://presagio.pages.dev/markets?id=0x0129d8d46e5869D10956BECf8E4A793309B04909

It's special because finalization period was only 1 hour.

On Presagio, it's still shown in Finalizing filter:

Screenshot by Dropbox Capture

And in the detail it also says I can still challenge it:

Screenshot by Dropbox Capture

This is because there is hard-coded 24h finalization period in the code, this PR is fixing that.  🙏 

On localhost in this branch:

Screenshot by Dropbox Capture

Screenshot by Dropbox Capture

(this is bugged on this special market only atm, but we are going to change finalization period from 1 day to 3 days and then it would be bugged everywhere: https://github.com/gnosis/prediction-market-agent-tooling/pull/466)

### DIFF
--- a/app/filters.tsx
+++ b/app/filters.tsx
@@ -4,7 +4,7 @@ import {
   OrderDirection,
 } from '@/queries/omen';
 import { AI_AGENTS_ALLOWLIST } from '../constants';
-import { _24HoursInSeconds, nowTimestamp } from '@/utils/time';
+import { nowTimestamp } from '@/utils/time';
 import {WXDAI, SDAI} from '@/constants/index';
 
 export type OrderFilter = {
@@ -53,8 +53,6 @@ export type StateFilter = {
   when: FixedProductMarketMaker_Filter;
 };
 
-const oneDayAgoTimestamp = nowTimestamp - _24HoursInSeconds;
-
 export const stateFilters: StateFilter[] = [
   {
     name: 'Open',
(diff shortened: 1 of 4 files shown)


kongzii:
This PR also closes https://github.com/SwaprHQ/presagio/issues/101

berteotti:
Thanks for the fix @kongzii !

Although we shouldn't have 24h constant value, I think the right solution is to fetch each market finalization period and check if answer finalized timestamp + finalizing period is already in the past (I don't know how to fetch that data).

Without checking for finalizing period we show markets that are still finalizing (the final answer can be disputed in reality.eth) as closed.

What do you think @kongzii ?

kongzii:
Hey @berteotti, I think there is misunderstanding of how `answerFinalizedTimestamp` works. It's the time when the market (the last answer on Reality) is finalized (so you don't need to add a finalization period to it; it's already included), and it's updated after each new answer.

We can test it out, let's take this question: https://reality.eth.limo/app/#!/question/0x79e32ae03fb27b07c89c0c568f80287c01ca2e57-0xebd865749abbe9e9576dbb2fcb2d744cb4bfd37f292142cb9bb17810ae9f9ab1

Screenshot by Dropbox Capture

I just answered it, deadline is 23 hours, subgraph returns  `answerFinalizedTimestamp = 1728122230`, which is `GMT: Saturday 5 October 2024 09:57:10`:

Screenshot by Dropbox Capture

If I answer it again:

Screenshot by Dropbox Capture

Timestamp is updated to `1728122455`, which is `GMT: Saturday 5 October 2024 10:00:55`

Screenshot by Dropbox Capture

berteotti (approved):

berteotti:
Hey @kongzii , didn't know about that field in the subgraph. Looks like it is working better than the previous solution. Awesome that we can get rid of the 24 hour constant. Thanks for the improvements!

PR #116, opened 2024-10-15
update @swapr/ui lib to v0.1.4

Diogomartf:
**Add new code icon**
image

**Latest screenshots**
image
image
image

**accommodate styles to new `Dialog`**
image
image
image
**mobile**
image
image

**some problems on the "old"**
image
**bad padding x space**
image


berteotti:
We should wait for the new release to fix the dialog close button position, right?

Diogomartf:
> We should wait for the new release to fix the dialog close button position, right?

No, we don't have to wait, will just need to update the package, nothing else.

Diogomartf:
> We should wait for the new release to fix the dialog close button position, right?

Updated!

package.json
```diff
@@ -13,7 +13,7 @@
     "@duneanalytics/client-sdk": "^0.2.1",
     "@fvictorio/newton-raphson-method": "^1.0.5",
     "@radix-ui/react-scroll-area": "^1.0.5",
-    "@swapr/ui": "^0.1.3",
+    "@swapr/ui": "^0.1.41",
```
berteotti, line 16: Why is it 0.1.41? 😅 Shouldn't be 0.1.4 ?

berteotti (approved):

ElRodrigote (approved):
Love it!

package.json
Diogomartf: Nope, it's the newest version.

PR #120, opened 2024-10-16
add Lifi widget to get sDAI with Popover (only desktop for now)

Diogomartf:
Makes it easy for people to get sDAI token and bet on presagio.

We are out of space on the top navbar on mobile, it's out of scope of this PR to fix that. 
This can work with `popover` out of the box and feels nicer than modal on desktop.

**screenshots with Popover**
image

**screenshots modal (previous)**
image
image


app/(main)/layout.tsx
```diff
@@ -1,4 +1,6 @@
 import { Footer, Navbar } from '@/app/components';
+import { LifiWidget } from '@/app/components/LifiWidget';
```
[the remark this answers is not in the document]
Diogomartf: fixed.

app/(main)/layout.tsx
```diff
@@ -14,6 +15,7 @@ export default function MainLayout({
           <Navbar />
           {children}
           <Footer />
+          <LifiWidgetModal />
```
berteotti, line 18: Why are we declaring this here? Why not inside Navbar? I think we should declare where it's being used to be easier to find and understand.

ElRodrigote: Like the settings popover? I like this suggestion

Diogomartf: Yeah, it's better

ElRodrigote (approved):

app/components/LifiWidgetPopover.tsx
```diff
@@ -0,0 +1,22 @@
...
+  return (
+    <Popover>
+      <PopoverTrigger asChild>
+        <Button variant="pastel" className="space-x-2 text-nowrap" onClick={() => {}}>
+          <p>Get {SDAI.symbol}</p>
+          <TokenLogo address={SDAI.address} size="xs" className="hidden md:block" />
+        </Button>
+      </PopoverTrigger>
+      <PopoverContent className="max-w-80 px-4 md:max-w-lg">
+        <LifiWidget />
```
berteotti, line 18: Should we delete this component and have the logic here? I think in this case there is not so much code, so it's good to have everything in 1 file.

Diogomartf: yeah, that's a good idea.

berteotti (approved):

app/components/LifiWidgetPopover.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <Popover>
+      <PopoverTrigger asChild>
+        <Button variant="pastel" className="space-x-2 text-nowrap">
+          <p>Get {SDAI.symbol}</p>
+        </Button>
+      </PopoverTrigger>
+      <PopoverContent className="max-w-80 px-4 md:max-w-lg">
+        <LiFiWidget integrator="Presagio" config={widgetConfig} />
```
ElRodrigote, line 42:
```suggestion
        <LiFiWidget integrator={APP_NAME} config={widgetConfig} />
```

ElRodrigote (approved):

app/components/LifiWidgetPopover.tsx
Diogomartf: yes

Issue #104, opened 2024-09-29
`canRedeem` logic incorrectly assuming you can only redeem if winner, ignoring invalid case


greenlucid:
Reality can return invalid. The reality proxy that generates the payouts handles this case:

https://gnosisscan.io/address/0xab16d643ba051c11962da645f74632d3130c81e2#code

<img width="675" alt="image" src="https://github.com/user-attachments/assets/1d98e068-f916-49d6-b5f5-80b7a26260ea">

The questions created with Kleros arbitrator, can be arbitrated according to this policy, which decides what Invalid means. Invalid is a valid case, so reality can and will return it.

https://cdn.kleros.link/ipfs/QmZM12kkguXFk2C94ykrKpambt4iUVKsVsxGxDEdLS68ws/omen-rules.pdf

Two components that allow the user to redeem (getting them from #103 PR branch):

https://github.com/SwaprHQ/presagio/blob/abd67565d92415c4b6d3bfe5048f797c2432f9d0/app/components/CardBet.tsx
https://github.com/SwaprHQ/presagio/blob/abd67565d92415c4b6d3bfe5048f797c2432f9d0/app/components/UserBets.tsx

Which ultimately use a few "entities" helpers, one of them leads us to the issue:

https://github.com/SwaprHQ/presagio/blob/abd67565d92415c4b6d3bfe5048f797c2432f9d0/entities/markets/marketCondition.ts#L17

`canRedeem` is tied to `isWinner` which is false. If the market is "invalid", all options will have the same payout, so if you have any shares at all, you can redeem.

0xNF21:
I am a user, and this prediction has been finalized as invalid, so I cannot recover my money through Presagio. Is there something we can do about it?

https://presagio.pages.dev/markets?id=0x888209942881afb58f1bf4f0575094ca7eed6022

PR #122, opened 2024-10-18
feat: allow users to redeem positions from invalid markets

berteotti:
image

Closes: #104

### DIFF
--- a/app/components/CardBet.tsx
+++ b/app/components/CardBet.tsx
@@ -14,6 +14,7 @@ import {
   Position,
   tradesCollateralAmountSpent,
   tradesOutcomeBalance,
+  outcomeTokensTradedTotal,
   UserBets,
 } from '@/entities';
 import { redeemPositions } from '@/hooks/contracts';
@@ -31,8 +32,12 @@ export const CardBet = ({ userBets, children }: CardBetProps) => {
   const outcomeIndex = position.getOutcomeIndex();
 
   const market = new Market(userBets.fpmm);
+  const marketCondition = new MarketCondition(userBets.fpmm, position.condition);
 
   const collateralAmountSpent = tradesCollateralAmountSpent({
+    fpmmTrades: userBets?.fpmmTrades,
+  });
+  const outcomeTokensTraded = outcomeTokensTradedTotal({
     fpmmTrades: userBets?.fpmmTrades,
   });
   const lastTradeTimestamp =
@@ -47,11 +52,31 @@ export const CardBet = ({ userBets, children }: CardBetProps) => {
   const isWinner = market.isWinner(outcomeIndex);
   const isLoser = market.isLoser(outcomeIndex);
 
-  const outcomeAmountString = market.isClosed
-    ? isWinner
-      ? 'Won'
-      : 'Lost'
-    : 'Potential win';
+  const getResultAmountString = (): string => {
+    if (!market.isClosed) return 'Potential win';
+    if (isWinner) return 'Won';
+    if (market.isAnswerInvalid) return 'Receive';
+
+    return 'Lost';
+  };
+
+  const getResultAmount = (): string => {
+    const invalidMarketColleteralToRedeem = marketCondition.getRedeemableColleteralToken(
+      outcomeTokensTraded,
+      outcomeIndex
+    );
+
+    if (!market.isClosed || isWinner)
+      return formatValueWithFixedDecimals(outcomeBalance, 2);
+
+    if (market.isAnswerInvalid)
+      return formatValueWithFixedDecimals(invalidMarketColleteralToRedeem || 0, 2);
+
+    return formatEtherWithFixedDecimals(collateralAmountSpent, 2);
+  };
+
+  const resultAmountString = getResultAmountString();
+  const resultAmount = getResultAmount();
 
   return (
     <Card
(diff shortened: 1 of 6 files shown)


app/components/CardBet.tsx
```diff
@@ -104,11 +114,13 @@ export const CardBet = ({ userBets, children }: CardBetProps) => {
...
-                  <p>{formatValueWithFixedDecimals(outcomeBalance)}</p>
-                ) : (
-                  <p>{formatEtherWithFixedDecimals(collateralAmountSpent, 2)}</p>
-                )}
+                <p>
+                  {!market.isClosed || isWinner
+                    ? formatValueWithFixedDecimals(outcomeBalance)
+                    : market.isAnswerInvalid
+                      ? formatValueWithFixedDecimals(invalidMarketColleteralToRedeem, 2)
+                      : formatEtherWithFixedDecimals(collateralAmountSpent, 2)}
```
Diogomartf, line 122: I think it's happening too much here, I'm okay with letting this here for now but we should find a better place for this logic than the the **jsx**

entities/markets/trade.ts
```diff
@@ -36,6 +36,15 @@ export const tradesCollateralAmountSpent = ({ fpmmTrades }: TradesProps) => {
...
 
+export const tradesOutcomeTokensTraded = ({ fpmmTrades }: TradesProps) => {
+  return (
+    fpmmTrades?.reduce((acc, trade) => {
+      const type = trade.type;
+      return valueByTradeBigInt[type](acc, BigInt(trade.outcomeTokensTraded));
+    }, BigInt(0)) ?? BigInt(0)
+  );
+};
+
```
Diogomartf, line 47: I don't think `tradesOutcomeTokensTraded` is a good name.

It return the total, right? Not the array of trades. So I think we should add `total` or `sum` to the name.

Maybe something like `outcomeTokensTradedTotal` would describe better

app/components/UserBets.tsx
```diff
@@ -202,16 +221,21 @@ export const UserBets = ({ fixedProductMarketMaker }: UserBets) => {
                         )}
                       >
                         <span>
-                          {tradedBalance} {collateralToken.symbol}
+                          {marketModel.isAnswerInvalid
+                            ? invalidMarketColleteralToRedeem
+                            : !canClaim && isResolved
+                              ? collateralSpent
+                              : tradedBalance}{' '}
+                          {collateralToken.symbol}
```
ElRodrigote, line 229: [MINOR] [NON-BLOCKING]

I have the same opinionated and personal take here, feel free to use or dismiss the suggestion:

```tsx
function getResultAmount() {
  if (marketModel.isAnswerInvalid) return invalidMarketColleteralToRedeem;
  if (!canClaim && isResolved) return collateralSpent;

  return tradedBalance;
}

const resultAmount = getResultAmount();
```

And then inside the TSX side of the code we could do:
```tsx
<span>
  {resultAmount} {collateralToken.symbol}
</span>
```

app/components/UserBets.tsx
```diff
@@ -239,6 +263,34 @@ export const UserBets = ({ fixedProductMarketMaker }: UserBets) => {
                     </div>
                   </>
                 )}
+                {marketModel.isAnswerInvalid && canRedeem && (
+                  <>
```
ElRodrigote, line 267: [MINOR] [NON-BLOCKING]

I think it's safe to remove this fragment

app/components/CardBet.tsx
```diff
@@ -47,12 +52,24 @@ export const CardBet = ({ userBets, children }: CardBetProps) => {
...
+  const invalidMarketColleteralToRedeem =
+    parseFloat(formatEther(outcomeTokensTraded)) * 0.5;
+
+  const resultAmount =
+    !market.isClosed || isWinner
+      ? formatValueWithFixedDecimals(outcomeBalance, 2)
+      : market.isAnswerInvalid
+        ? formatValueWithFixedDecimals(invalidMarketColleteralToRedeem, 2)
+        : formatEtherWithFixedDecimals(collateralAmountSpent, 2);
+
```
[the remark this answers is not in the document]
Diogomartf: Good suggestion, I think something like this is a step in the right direction.

greenlucid:
Wording on the bottom is wrong, isn't it? If spent on bet was 30 WXDAI, but they receive 16.295 WXDAI from redeeming, they they got 32.59 shares, not 30 shares.
Also an user might have both YES and NO shares in the same market, no? Does that affect the rendering if that was the case?

berteotti:
> Wording on the bottom is wrong, isn't it? If spent on bet was 30 WXDAI, but they receive 16.295 WXDAI from redeeming, they they got 32.59 shares, not 30 shares. 

You are right, user has 32.59 shares from 30 WXDAI he spent. Thanks for noticing.

> Also an user might have both YES and NO shares in the same market, no? Does that affect the rendering if that was the case?

We show each bet outcome independently.  So if you bet on YES and NO, we show 2 separate bets, showing share amount for each outcome.

app/components/CardBet.tsx
```diff
@@ -14,13 +14,15 @@ import {
...
   tradesOutcomeBalance,
+  outcomeTokensTradedTotal,
   UserBets,
 } from '@/entities';
 import { redeemPositions } from '@/hooks/contracts';
 import { MarketThumbnail } from './MarketThumbnail';
 import { Skeleton } from './Skeleton';
 import { formatEtherWithFixedDecimals, formatValueWithFixedDecimals } from '@/utils';
 import { useTx } from '@/context';
+import { formatEther } from 'viem';
```
ElRodrigote, line 25: [MINOR][NON-BLOCKING]
We can safely remove this import.

```suggestion
```

ElRodrigote (approved):

Diogomartf (approved):

PR #123, opened 2024-10-22
Add new footer

Diogomartf:
Add a better footer

**features**
- show partners that help power Presagio
- show socials
- show market categories

**Screenshots**
image

image

### DIFF
--- a/app/components/Footer.tsx
+++ b/app/components/Footer.tsx
@@ -1,10 +1,164 @@
+import { APP_NAME, Categories, DISCORD_URL, GITHUB_URL, X_URL } from '@/constants';
+import Link from 'next/link';
+
 export const Footer = () => (
-  <div className="flex items-center justify-center space-x-3 px-6 pt-32">
-    <p>
-      built by{' '}
-      <a href="https://swapr.eth.limo" className="text-text-primary-main">
-        Swapr.eth
-      </a>
-    </p>
+  <div className="mt-32 border-t border-surface-surface-2 bg-surface-surface-0 px-6 md:px-10 lg:px-20 xl:px-40">
+    <div className="flex flex-col-reverse items-start justify-between py-10 md:flex-row">
+      <div className="w-full space-y-12 md:w-fit">
+        <div className="hidden space-y-2 md:block">
+          <Logo />
+          <p>Where prediction markets and AI agents meet</p>
+        </div>
+        <PoweredByPartners />
+      </div>
+      <div className="flex space-x-20">
+        <div>
+          <p className="mb-2 text-lg font-bold">Markets</p>
+          <ul className="space-y-1">
+            {Object.values(Categories).map(category => (
+              <li className="text-md font-medium capitalize" key={category}>
+                <a href={`/?c=${category}`}>
+                  <p className="hover:text-text-primary-main">{category}</p>
+                </a>
+              </li>
+            ))}
+          </ul>
+        </div>
+        <div>
+          <p className="mb-2 text-lg font-bold">Connect</p>
+          <ul className="space-y-1">
+            <li className="text-md capitalize">
+              <a href={DISCORD_URL} target="_blank">
+                <p className="hover:text-text-primary-main">Discord</p>
+              </a>
+            </li>
+            <li className="text-md capitalize">
+              <a href={GITHUB_URL} target="_blank">
+                <p className="hover:text-text-primary-main">Github</p>
+              </a>
+            </li>
+            <li className="text-md capitalize">
+              <a href={X_URL} target="_blank">
+                <p className="hover:text-text-primary-main">X</p>
+              </a>
+            </li>
+          </ul>
+        </div>
+      </div>
+    </div>
+    <MadeBySwapr />
   </div>
+);
+
+const PoweredByPartners = () => (
+  <div className="flex flex-wrap items-center gap-2 rounded-8 bg-surface-surface-1 p-3">
+    <p className="">Powered by</p>
+    <a
+      href="https://www.gnosis.io/"
+      className="flex items-center space-x-1"
+      target="_blank"
+    >
+      <GnosisIcon />
+      <p className="font-bold">Gnosis</p>
+    </a>
+    <span className="">•</span>
+    <a href="https://reality.eth.limo/" className="font-bold" target="_blank">
+      Reality.eth
+    </a>
+    <span className="">•</span>
+    <a href="https://kleros.io/" className="flex items-center space-x-1" target="_blank">
+      <KlerosIcon />
+      <p className="font-bold">Kleros</p>
+    </a>
+    <span className="">•</span>
+    <a href="https://www.gnosis.io/labs" className="font-bold" target="_blank">
+      Gnosis AI
+    </a>
+  </div>
+);
+
+const MadeBySwapr = () => (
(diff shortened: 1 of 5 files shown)


berteotti:
I think we can have the "Powered by" component extend full width and break if there isn't more space. Also we can have more bottom padding so it's easier to read "Made by". See image example.

image

app/components/Footer.tsx
```diff
@@ -1,10 +1,165 @@
...
+const Logo = () => (
+  <Link href="/">
+    <div className="text-white flex items-center text-[24px] font-black md:space-x-2">
+      <p>👁️</p>
+      <p>{APP_NAME}</p>
+    </div>
+  </Link>
+);
+
+const SwaprIcon = () => (
```
ElRodrigote, line 108: I'd move the 3 icons (Swapr, Gnosis and Kleros) to a separate assets folder, and not have them here.

app/components/Footer.tsx
```diff
@@ -1,10 +1,165 @@
...
+      </div>
+    </div>
+    <MadeBySwapr />
+  </div>
+);
+
+const PoweredByPartners = () => (
+  <div className="flex flex-col space-y-2 rounded-8 bg-surface-surface-1 p-3 md:flex-row md:items-center md:space-x-2 md:space-y-0">
+    <p className="mb-2 text-nowrap md:mb-0 md:mr-2">Powered by</p>
+    <div className="grid grid-cols-2 gap-2 md:flex">
```
ElRodrigote, line 51: Ahhh love this change, I like this better than the single column <3, great work!

app/components/Footer.tsx
```diff
@@ -1,10 +1,165 @@
...
+    <div className="flex flex-col-reverse items-start justify-between py-10 md:flex-row">
+      <div className="w-full space-y-12 md:w-fit">
+        <div className="hidden space-y-2 md:block">
+          <Logo />
+          <p>Where prediction markets and AI agents meet</p>
+        </div>
+        <PoweredByPartners />
+      </div>
+      <div className="flex space-x-10">
+        <div>
```
ElRodrigote, line 15: I don't know about placing the Markets Categories in the footer, it feels out of place.

I like the socials, some other ideas we could place in the footer IMHO are:
* The license this product is under
* a link to our GitHub
* maybe some contact info outside the socials

If we want to focus on the markets, I'd place a floating icon button with a caret pointing up instead of the Markets. This button, on click, would take the user to the page start. As we don't have a very complex sitemap structure I wouldn't place much more links here.

app/components/Footer.tsx
```diff
@@ -1,10 +1,165 @@
...
+      <span className="hidden md:inline-block">•</span>
+      <a href="https://www.gnosis.io/labs" className="font-bold" target="_blank">
+        Gnosis AI
+      </a>
+    </div>
+  </div>
+);
+
+const MadeBySwapr = () => (
+  <div className="mt-12 flex items-center justify-center border-t border-surface-surface-2 py-5 md:justify-start">
```
ElRodrigote, line 82: [MINOR][NON-BLOCKING][OPINIONATED]:
I feel there's a lot of blank space between the footer content and this component, I think the padding is enough here. Some visual examples:

```suggestion
  <div className="flex items-center justify-center border-t border-surface-surface-2 py-5 md:justify-start">
```

### Now (desktop):
Screenshot 2024-10-22 at 11 24 38
### Without margin (desktop):
Screenshot 2024-10-22 at 11 24 48
-----------

### Now (mobile):
Screenshot 2024-10-22 at 11 26 10
### Without margin (mobile):
Screenshot 2024-10-22 at 11 26 17

ElRodrigote (approved):

app/components/Footer.tsx
Diogomartf: On spacing thing, agree it's a bit too much, but I'll keep some space to breath, I'll settle in half: `mt-6`.

app/components/Footer.tsx
Diogomartf: Not sure about this, as I we probably will not use them anywhere else. Being on this file I agree it makes more complex but keeps complexity here instead and does not add more files to the project.

@berteotti what's your thoughts on this?

ElRodrigote (approved):

app/components/Footer.tsx
berteotti: Agree on adding complexity in this file only instead of adding complexity to the codebase. If we need this in more than 1 file, let's move it. Otherwise it can live here.

app/components/MadeBySwapr.tsx
berteotti, line 1: Since this is only used in the footer, probably it's a good thing to keep it on `Footer.tsx`, as the icons. IMO, this will make the codebase more readable with less components files. We should make the rule that components must be used twice or more to be in this folder. What do you think?

app/components/MadeBySwapr.tsx
```diff
@@ -0,0 +1,86 @@
+'use client';
+
+import { useState, useEffect } from 'react';
+
+const cities = ['Stockholm', 'Porto', 'Barcelona'];
+
+const getRandomCityIndex = () => Math.floor(Math.random() * cities.length);
```
berteotti, line 7: Why random? Probably a loop would be better since random can show the same city for a long time or just show 2 of them.

app/components/MadeBySwapr.tsx
```diff
@@ -0,0 +1,86 @@
...
+    const updateCity = () => {
+      setIsAnimating(true);
+
+      setTimeout(() => {
+        setCurrentCity(getRandomCity());
+        setIsAnimating(false);
+      }, 200);
+    };
+
+    const DURATION_IN_MS = 10000; // 10 sec
```
berteotti, line 33: Why not call the variable `_10_SECONDS`, `TEN_SECONDS` or `_10_SECONDS_IN_MS` and delete the comment?

berteotti (commented):
I think it's almost ready to merge. Just having a few concerns.

app/components/MadeBySwapr.tsx
berteotti: I think it's ok like it is. Just want us to be aware and think about this, and to think how we can have a simpler codebase. It is super easy/quick for the codebase to get complex.

app/components/MadeBySwapr.tsx
Diogomartf: true, that's an issue, the idea was be more dynamic, but maybe it's better sequentially.

app/components/MadeBySwapr.tsx
ElRodrigote: > Since this is only used in the footer, probably it's a good thing to keep it on `Footer.tsx`, as the icons. IMO, this will make the codebase more readable with less components files. We should make the rule that components must be used twice or more to be in this folder. What do you think?

I agree with this. For readability purposes, if only used once we could use the same file to define the satellite components

Diogomartf: I think that's a good rule of thumb to decide to move, but there are cases where it's best to separate the component (eg:  complex components, file getting too big). So not a de facto rule but a good thinking point.

ElRodrigote (approved):

Diogomartf:
> I think we can use some css instead of js for the animation. But it's good enough

You are right. Updated with css only approach.

berteotti (approved):
Good job Diogo!

PR #125, opened 2024-10-23
feat: improve tx creation experience

berteotti:
image

https://github.com/user-attachments/assets/4d0b31b1-b5e9-4917-a4f8-e7ce13572d73

### DIFF
--- a/app/components/ConfirmTrade.tsx
+++ b/app/components/ConfirmTrade.tsx
@@ -8,7 +8,10 @@ import {
   DialogFooter,
   DialogHeader,
   DialogTitle,
+  errorToast,
   Icon,
+  successToast,
+  toast,
 } from '@swapr/ui';
 import { SLIPPAGE, SwapDirection, SwapState } from '.';
 import MarketABI from '@/abi/market.json';
@@ -20,8 +23,17 @@ import ConditionalTokensABI from '@/abi/conditionalTokens.json';
 import { addFraction, removeFraction } from '@/utils/price';
 import { Abi, Address, erc20Abi, formatEther, parseEther } from 'viem';
 import { useTx } from '@/context';
-import { Token } from '@/entities';
-import { formatEtherWithFixedDecimals } from '@/utils';
+import { Outcome, Token } from '@/entities';
+import { formatEtherWithFixedDecimals, shortenAddress } from '@/utils';
+import {
+  waitForTransactionReceipt,
+  WaitForTransactionReceiptErrorType,
+  writeContract,
+  WriteContractErrorType,
+} from 'wagmi/actions';
+import { config } from '@/providers/chain-config';
+import { SVGProps, useEffect, useState } from 'react';
+import { GNOSIS_SCAN_URL } from '@/constants';
 
 const ROUNDING_PRECISON = 0.00000000001;
 
@@ -48,6 +60,13 @@ export const ConfirmTrade = ({
 }: ConfirmTradeProps) => {
   const { isModalOpen, closeModal } = useModal();
   const { submitTx } = useTx();
+  const [isApproving, setIsApproving] = useState(false);
+
+  useEffect(() => {
+    if (!swapState.isAllowed) {
+      setIsApproving(false);
+    }
+  }, [swapState.isAllowed]);
 
   const { data: sellAmount } = useReadCalcSellAmount(
     marketId,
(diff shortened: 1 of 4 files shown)


Diogomartf:
The icon on the toast is not correct on success, we should update it to a check.
Also it is very small, not sure what's causing this.

image

Diogomartf:
The experience is much better!! Using toast for approvals is indeed much better! Good job.

app/layout.tsx
```diff
@@ -13,7 +13,7 @@ const manrope = Manrope({ subsets: ['latin'] });
 export const metadata: Metadata = {
   metadataBase: new URL(APP_URL),
   title: `${APP_NAME} - Where predictions meet AI agents`,
-  description: 'Bet now on multiple markets in gnosis chain.',
+  description: 'Bet on prediciton markets in Gnosis Chain, powered by AI.',
```
Diogomartf, line 16: powered by AI is a bit misleading, since it's not powered by AI, It's just AI aware at this point (showing AI agents). We'll show research, and models used by agents to explain the actions.

```suggestion
  description: 'Bet on prediciton markets in Gnosis Chain and follow AI agents bets and reasoning.',
```

Something like this.

providers/providers.tsx
```diff
@@ -23,7 +23,10 @@ export const Providers = ({ children }: PropsWithChildren) => {
           <CustomConnectKitProvider>
             <ModalProvider>
               <TxProvider>
-                <TooltipProvider delayDuration={300}>{children}</TooltipProvider>
+                <TooltipProvider delayDuration={300}>
+                  {children}
+                  <Toaster position="top-right" toastOptions={{ duration: 7000 }} />
+                </TooltipProvider>
```
Diogomartf, line 29: Should we move this provider above, next to modal, seems that they are quite similar, would like them to be together.

app/components/ConfirmTrade.tsx
```diff
@@ -65,27 +77,92 @@ export const ConfirmTrade = ({
...
 
+  const waitingTxToast = (txHash: string) =>
+    toast({
+      children: (
+        <div className="flex items-center space-x-4">
+          <Spinner className="h-5 w-5 shrink-0 animate-spin" />
+          <div>
+            Wating for transaction success{' '}
+            <a
+              href={`https://gnosisscan.io/tx/${txHash}`}
```
Diogomartf, line 88: we have a GNOSIS_SCAN_URL constant, let's use that, we could probably have a function to create a `blockExplorerTxLink`.

And maybe GNOSIS_SCAN_URL -> BLOCK_EXPLORER_URL would be better, cause we are not tied to gnosis scan, we can even choose to use other like blockscout.com at some point.

app/components/ConfirmTrade.tsx
```diff
@@ -65,27 +77,92 @@ export const ConfirmTrade = ({
...
+              target="_blank"
+              className="inline-block underline"
+              rel="noopener noreferrer"
```
Diogomartf, line 91:
```suggestion
```
I think it's safe to remove this `rel`, I think we already had a discussion where we found out that this is not a vulnerability anymore.

app/components/ConfirmTrade.tsx
```diff
@@ -200,3 +284,21 @@ export const ConfirmTrade = ({
...
+      stroke="currentColor"
+      stroke-width="4"
+    ></circle>
+    <path
+      className="opacity-75"
+      fill="currentColor"
+      d="M4 12a8 8 0 018-8V0C5.373 0 0 5.373 0 12h4zm2 5.291A7.962 7.962 0 014 12H0c0 3.042 1.135 5.824 3 7.938l3-2.647z"
+    ></path>
+  </svg>
+);
```
Diogomartf, line 304: 1. Should we add this icon to our icon set on `swapr-ui`?

2. This is a good example of a  component I'm in favor of moving this spinner out to it's own file. why? 
Because we'll probably will use some kind of loading in the future, and I'm afraid, by being hidden inside this file, we don't remember we already have a spinner.

app/layout.tsx
berteotti: We are also showing markets created with AI. What about adding some mystery to what AI does so users click on it?

berteotti:
> The icon on the toast is not correct on success, we should update it to a check. Also it is very small, not sure what's causing this.


This has to be fixed in swapr-ui.

app/components/ConfirmTrade.tsx
berteotti: I think we should add it to `swapr-ui`. But in the meantime, while we don't add it I think it can live here.

providers/providers.tsx
berteotti: It's kind of the same but did it anyways, let me know how it looks.

app/layout.tsx
Diogomartf: I would prefer something like `enhanced by AI`, powered seems to strong for this case.

app/components/ConfirmTrade.tsx
Diogomartf: can you create a ticket to add spinner icon to `swapr-ui`?

It can live here in the meanwhile.

Diogomartf:
> > The icon on the toast is not correct on success, we should update it to a check. Also it is very small, not sure what's causing this.
> 
> This has to be fixed in swapr-ui.

Which part isn't supported, the check? the size? both?

We need to create a ticker for fixing this then.

berteotti:
> Which part isn't supported, the check? the size? both?

Both

> We need to create a ticker for fixing this then.

Here it is https://linear.app/swaprhq/issue/SWUI-45/fix-status-toast-icon

app/components/ConfirmTrade.tsx
berteotti: Here it is https://linear.app/swaprhq/issue/SWUI-44/add-spinner-icon

app/components/ConfirmTrade.tsx
berteotti: Created a task for it https://linear.app/swaprhq/issue/PGIO-119/create-function-to-return-explorer-link

Diogomartf (approved):

app/layout.tsx
```diff
@@ -13,7 +13,7 @@ const manrope = Manrope({ subsets: ['latin'] });
 export const metadata: Metadata = {
   metadataBase: new URL(APP_URL),
   title: `${APP_NAME} - Where predictions meet AI agents`,
-  description: 'Bet now on multiple markets in gnosis chain.',
+  description: 'Bet on prediciton markets in Gnosis Chain, enhanced by AI.',
```
ElRodrigote, line 16: Small typo

```suggestion
  description: 'Bet on prediction markets in Gnosis Chain, enhanced by AI.',
```

app/components/ConfirmTrade.tsx
```diff
@@ -65,27 +78,90 @@ export const ConfirmTrade = ({
...
-    });
+    })
+      .then(async txHash => {
+        waitingTxToast(txHash);
+
+        await waitForTransactionReceipt(config, {
+          hash: txHash,
+        });
+
+        succesTxToast(txHash, inToken);
```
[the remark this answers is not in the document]
berteotti: The tx can fail (not cancelled), right? If it fails, it will be caught by `catch()`  and the user will have to submit a new tx. Probably we should show an error toast.

app/components/ConfirmTrade.tsx
```diff
@@ -65,27 +78,90 @@ export const ConfirmTrade = ({
     ? formatEtherWithFixedDecimals(tokenAmountOut)
     : '';
 
+  const waitingTxToast = (txHash: string) =>
+    toast({
+      children: (
+        <div className="flex items-center space-x-4">
+          <Spinner className="h-5 w-5 shrink-0 animate-spin" />
+          <div>
+            Wating for transaction success{' '}
```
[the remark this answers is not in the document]
berteotti: I think submission is a better term here. Because the user wants to submit the transaction in a block. It can succeed, it can fail.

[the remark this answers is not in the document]
berteotti: Maybe confirmation is better. Explores use this term.

app/components/ConfirmTrade.tsx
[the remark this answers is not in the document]
berteotti: Done, thanks for the feedback!

PR #128, opened 2024-10-23
filter "test" market category when no category is passed

Diogomartf:
This will enable the agents to create a test category meant for testing markets that will not appear on all markets but can still be filtered.

### DIFF
--- a/queries/omen/markets.ts
+++ b/queries/omen/markets.ts
@@ -179,6 +179,7 @@ const getMarketsQuery = (
         ${params.resolutionTimestamp !== undefined ? 'resolutionTimestamp: $resolutionTimestamp' : ''}
         ${params.currentAnswerTimestamp_gt !== undefined ? 'currentAnswerTimestamp_gt: $currentAnswerTimestamp_gt' : ''}
         ${params.collateralToken_in !== undefined ? 'collateralToken_in: $collateralToken_in' : ''}
+        ${!params.category_contains ? 'category_not_contains: "test"' : ''}
       }
     ) {
       ...marketData


queries/omen/markets.ts
```diff
@@ -166,7 +166,7 @@ const getMarketsQuery = (
         ${params.title_contains_nocase ? 'title_contains_nocase: $title_contains_nocase' : ''}
         ${params.id ? 'id: $id' : ''}
         ${params.creator_in ? 'creator_in: $creator_in' : ''}
-        ${params.category_contains ? 'category_contains: $category_contains' : ''}
+        ${params.category_contains ? 'category_contains: $category_contains' : 'category_not_contains: "test"'}
```
berteotti, line 169: I think we should reserve this line only for `category_contains`.

Let's add `category_not_contains` in another line.

Will simplify how we read too.

queries/omen/markets.ts
```diff
@@ -179,6 +179,7 @@ const getMarketsQuery = (
         ${params.resolutionTimestamp !== undefined ? 'resolutionTimestamp: $resolutionTimestamp' : ''}
         ${params.currentAnswerTimestamp_gt !== undefined ? 'currentAnswerTimestamp_gt: $currentAnswerTimestamp_gt' : ''}
         ${params.collateralToken_in !== undefined ? 'collateralToken_in: $collateralToken_in' : ''}
+        ${!params.category_contains ? 'category_not_contains: "test"' : ''}
```
berteotti, line 182: We can have it by default, right? Without any condition.

berteotti (approved):

PR #131, opened 2024-10-24
[FEAT][PGIO-114] - Analytics - track important events

ElRodrigote:
## Description
* Removes deprecated options in the Fathom load fn (see [this link](https://github.com/derrickreimer/fathom-client?tab=readme-ov-file#allowed-domains))
* Creates a SSOT to manage our FA event labeling
* Implements event tracking for interactions in/with:
  * Navbar "Get sDAI" button triggering the LiFi Widget 
  * Markets list: 
    * categories
    * search bar
    * filters
    * specific market selection
  * Market details:
    * details tabs and inner tabs selection
    * news article selection
    * info external links clicks to Reality, Gnosis Scan and Kleros
* [Pending]: 
  * [x] footer
  * [x] embed markets
  * [x] redeems
  * [x] assess any other pending important event. 

## Visual Evidence
Screenshot 2024-10-25 at 13 51 46


app/(main)/markets/News.tsx
```diff
@@ -77,6 +80,7 @@ const NewsArticle = ({ url, title }: NewsArticleProps) => {
       href={url}
       target="_blank"
       rel="noopener noreferrer"
+      onClick={() => trackEvent(FA_EVENTS.MARKETS.DETAILS.TABS.NEWS_ARTICLE(id))}
```
Diogomartf, line 83: I think here URL would be nicer

constants/index.ts
```diff
@@ -76,3 +76,61 @@ export const DISCORD_URL =
   process.env.NEXT_PUBLIC_DISCORD_URL || 'https://discord.com/invite/QFkNsjTkzD';
 export const GITHUB_URL =
   process.env.NEXT_PUBLIC_GITHUB_URL || 'https://github.com/SwaprHQ/presagio';
+
+const formatEventName = (eventName: string) =>
+  eventName.split(' ').join('-').toLocaleLowerCase();
+export const FA_EVENTS = {
+  FOOTER: {
+    CONNECT: {
+      DISCORD: 'click/footer-connect_discord',
```
berteotti, line 89: Why do we have dashes and underscores?

constants/index.ts
```diff
@@ -76,3 +76,61 @@ export const DISCORD_URL =
   process.env.NEXT_PUBLIC_DISCORD_URL || 'https://discord.com/invite/QFkNsjTkzD';
 export const GITHUB_URL =
   process.env.NEXT_PUBLIC_GITHUB_URL || 'https://github.com/SwaprHQ/presagio';
+
+const formatEventName = (eventName: string) =>
+  eventName.split(' ').join('-').toLocaleLowerCase();
+export const FA_EVENTS = {
```
berteotti, line 82: Probably it's better to have a analytics file, right?

constants/index.ts
ElRodrigote: I used them to express a hierarchy, in my head:

`$eventTrigger/$path-to-event_$event-detail`

constants/index.ts
ElRodrigote: Yes! I agree, this `FA_EVENTS` object feels out of place now hahaha, I'll move it, maybe to its own module (`@/analytics`) or inside the `useAnalytics` hook file

app/(main)/markets/News.tsx
```diff
@@ -47,11 +50,11 @@ export const News = ({ id }: NewsProps) => {
   if (!news) return <div className="py-4 text-center">No news</div>;
 
   return news.map((newsArticle, index) => (
-    <NewsArticle key={index} url={newsArticle.url} title={newsArticle.title} />
+    <NewsArticle id={id} key={index} url={newsArticle.url} title={newsArticle.title} />
   ));
 };
 
-const NewsArticle = ({ url, title }: NewsArticleProps) => {
+const NewsArticle = ({ id, url, title }: NewsArticleProps) => {
```
Diogomartf, line 57: Don't think we're using id anymore
```suggestion
const NewsArticle = ({ url, title }: NewsArticleProps) => {
```

app/(main)/markets/News.tsx
```diff
@@ -77,6 +80,7 @@ const NewsArticle = ({ url, title }: NewsArticleProps) => {
       href={url}
       target="_blank"
       rel="noopener noreferrer"
+      onClick={() => trackEvent(FA_EVENTS.MARKETS.DETAILS.TABS.NEWS_ARTICLE)}
```
Diogomartf, line 83: we can pass an article title tho

app/(main)/markets/News.tsx
```diff
@@ -47,11 +50,11 @@ export const News = ({ id }: NewsProps) => {
   if (!news) return <div className="py-4 text-center">No news</div>;
 
   return news.map((newsArticle, index) => (
-    <NewsArticle key={index} url={newsArticle.url} title={newsArticle.title} />
+    <NewsArticle id={id} key={index} url={newsArticle.url} title={newsArticle.title} />
```
Diogomartf, line 53:
```suggestion
    <NewsArticle key={index} url={newsArticle.url} title={newsArticle.title} />
```

analytics/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    FILTERS: {
+      CREATOR: (creatorName: string) =>
+        `click/markets/filters/creator-${formatEventName(creatorName)}`,
+      ORDER: (orderName: string) =>
+        `click/markets/filters/order-${formatEventName(orderName)}`,
+      STATE: (stateName: string) =>
+        `click/markets/filters/state-${formatEventName(stateName)}`,
+      TOKEN: (tokenName: string) =>
+        `click/markets/filters/token-${formatEventName(tokenName)}`,
+    },
```
berteotti, line 59: Maybe having a way to know what filters are applied on the url is better than tracking every click? Also, should we track every option click or just clicking on each filter select?

ElRodrigote: I didn't find a way to track search params from the URL yet. That would solve the market details tracking (as we could use the `?id` from the URL)

If we want to know if the users are using filters, we can drop the option clicks.
If we want to know the most popular filters, the current approach works.

Follows a similar heuristic to the one used for the market categories: registering a click on the categories control vs registering every category click.

app/(main)/markets/News.tsx
ElRodrigote: Yes! We can. I'm not sure if it's worth it to track which title the user clicks, as each market will have its own articles, so there will be a lot of single click events in market news that are closed already

analytics/index.ts
berteotti: We can go forward with this solution

berteotti (approved):
Good job Rorry, looks good to me!

PR #132, opened 2024-10-25
feat: show market status and winning outcome on cardbet

### LINKED ISSUE #126: Invalid market state not considered in `My Bets`
For example, this market https://presagio.pages.dev/markets?id=0x14d1c2c828ba13da17f1a38fb9b67ef7853ec711 is finalizing as Invalid.

Our rebel agent bet on `Yes` outcome because it was cheap (price <0.5), so after the market is resolved as invalid, he indeed will be profitable from this bet (because in the end, `Yes` token is worth 0.5 after market resolves as invalid).

However, in my bets section, it looks like the market doesn't have an answer yet and that he lost the whole investment:

Screenshot by Dropbox Capture

berteotti:
image

Example profile: https://feat-pgio-107-show-outcome-s.presagio.pages.dev/profile?address=0x91fEA319Ed45c7b12683c8c25e741151fBFee24b

Closes #126

### DIFF
--- a/app/(main)/markets/Bet.tsx
+++ b/app/(main)/markets/Bet.tsx
@@ -4,7 +4,6 @@ import {
   ButtonLink,
   Icon,
   Tag,
-  TagProps,
   Tooltip,
   TooltipContent,
   TooltipTrigger,
@@ -12,18 +11,12 @@ import {
 import { Market } from '@/entities';
 import { FixedProductMarketMaker } from '@/queries/omen';
 import { Swapbox } from '@/app/components';
-import { KLEROS_URL, REALITY_QUESTION_URL } from '@/constants';
+import { KLEROS_URL, OUTCOME_TAG_COLORS_SCHEME, REALITY_QUESTION_URL } from '@/constants';
 import { PropsWithChildren } from 'react';
 
 interface BetProps {
   fixedProductMarketMaker: FixedProductMarketMaker;
 }
-
-const tagSchemeColors: Record<number, TagProps['colorScheme']> = {
-  0: 'success',
-  1: 'danger',
-  [Market.INVALID_ANSWER]: 'quaternary',
-};
 
 export const Bet = ({ fixedProductMarketMaker }: BetProps) => {
   const marketModel = new Market(fixedProductMarketMaker);
@@ -82,7 +75,7 @@ export const Bet = ({ fixedProductMarketMaker }: BetProps) => {
           <span>Current winner outcome is</span>
           <Tag
             className="w-fit uppercase"
-            colorScheme={tagSchemeColors[marketModel.currentAnswer]}
+            colorScheme={OUTCOME_TAG_COLORS_SCHEME[marketModel.currentAnswer]}
           >
             {marketModel.currentAnswer === Market.INVALID_ANSWER
               ? 'Invalid'
@@ -123,7 +116,7 @@ export const Bet = ({ fixedProductMarketMaker }: BetProps) => {
         <div className="flex items-center space-x-2">
           <Tag
             className="w-fit uppercase"
-            colorScheme={tagSchemeColors[winnerOutcome.index]}
+            colorScheme={OUTCOME_TAG_COLORS_SCHEME[winnerOutcome.index]}
           >
             {winnerOutcome.name}
           </Tag>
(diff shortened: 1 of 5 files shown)


app/components/CardBet.tsx
```diff
@@ -83,42 +83,71 @@ export const CardBet = ({ userBets, children }: CardBetProps) => {
...
+              )}
             </div>
           </div>
         </section>
       </Link>
-      <section className="flex items-center border-t border-outline-base-em px-4 py-2 md:h-[48px] md:py-0">
+      <section className="flex items-center border-t border-outline-base-em p-4 md:h-[48px]">
         <div className="flex w-full items-center justify-between space-x-4">
-          <div className="flex flex-col items-start space-y-0.5 md:flex-row md:items-center md:space-x-3">
+          <div className="flex flex-col items-start space-y-1.5 md:flex-row md:items-center md:space-x-3 md:space-y-0">
```
Diogomartf, line 145: we probably can use gap here, so we don't need to md:space-y-0. 
Also I think we should increase the space x on md:, hence the `md:gap-5`

```suggestion
          <div className="flex flex-col items-start gap-y-1.5 md:flex-row md:items-center md:gap-5">
```

entities/markets/market.ts
```diff
@@ -68,8 +69,24 @@ export class Market {
...
-    return this.isClosed && this.answer !== null && this.answer !== Market.INVALID_ANSWER
+    return this.isClosed && this.answer !== null && !this.isAnswerInvalid
       ? this.outcomes[this.answer]
       : null;
   }
+
+  getCurrentAnswerOutcome() {
+    return this.currentAnswer !== null && !this.isCurrentAnswerInvalid
+      ? this.outcomes[this.currentAnswer]
+      : null;
```
Diogomartf, line 80: why we're checking if `isCurrentAnswerInvalid`, wouldn't we also expect to receive invalid if it is the case on `getCurrentAnswerOutcome()`?

Diogomartf: not sure I understood

berteotti: If answer is invalid we are giving it a -1 index. We can't get -1 index from outcomes array to represent an invalid outcome. Currently we are returning null. There is a possibility of using `.at` method to return `array.length - 1 ` outcome.

Diogomartf: yeah, we can try that

berteotti: Done

app/components/CardBet.tsx
berteotti: Used close values to ones you suggested

Diogomartf (approved):

app/(main)/markets/Bet.tsx
```diff
@@ -11,20 +11,14 @@ import {
...
-import { Swapbox } from '@/app/components';
+import { Swapbox, tagSchemeColors } from '@/app/components';
 import { KLEROS_URL, REALITY_QUESTION_URL } from '@/constants';
 import { PropsWithChildren } from 'react';
 
 interface BetProps {
   fixedProductMarketMaker: FixedProductMarketMaker;
 }
 
-const tagSchemeColors: Record<number, TagProps['colorScheme']> = {
```
ElRodrigote, line 22: We should remove the TagProps import

app/components/CardBet.tsx
```diff
@@ -27,6 +27,12 @@ interface CardBetProps extends PropsWithChildren {
   userBets: UserBets;
 }
 
+export const tagSchemeColors: Record<number, TagProps['colorScheme']> = {
+  0: 'success',
+  1: 'danger',
+  [Market.INVALID_ANSWER]: 'quaternary',
+};
```
ElRodrigote, line 34: Is it worth transforming this into a utils fn? 
We use a similar logic at:
`app/(main)/markets/MarketActivity.tsx`

The piece of code I'm talking about:
```tsx
const TAG_COLOR_SCHEMES: { 0: TagColorSchemeProp; 1: TagColorSchemeProp } = {
  0: 'success',
  1: 'danger',
};
```

ElRodrigote (approved):

app/components/CardBet.tsx
berteotti: I think an object is ok for this. Where should we declare it?

PR #134, opened 2024-10-28
show notification on my bets button when wallet has unredeemed bets

Diogomartf:
Show notification on "my bets" button when wallet has unredeemed bets to make it clear when user has to redeem.

**Preview**
image
image


entities/markets/userBets.ts
```diff
@@ -6,3 +8,32 @@ export interface UserBets extends UserPosition {
   fpmm: FixedProductMarketMaker;
   condition: Condition;
 }
+
+export class Bets {
```
berteotti, line 12: File is named UserBets but the class is named Bets, should we call it UserBets too? I think this is related to a user positions, so UserBets makes sense in my mind.

ElRodrigote: Yeah agree, and if in the future we add another feature like a ranking or sharing bets, or something like that, we can differentiate one Bet from another with the naming suggested by Leo.

app/components/Navbar.tsx
```diff
@@ -20,9 +41,16 @@ export const Navbar = () => {
...
-              My bets
-            </Button>
+            <div className="relative">
+              <Button variant="pastel" className="space-x-1 text-nowrap">
+                My bets
+              </Button>
+              {hasUnredeemedBets && (
+                <div className="absolute -right-2 -top-1 flex size-5 items-center justify-center rounded-100 border-2 border-surface-surface-bg bg-surface-success-main p-1 text-text-white">
+                  <p className="text-xs font-semibold">{unredeemedBetsNumber}</p>
+                </div>
```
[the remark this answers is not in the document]
Diogomartf: I read well, but can try black on dark mode

entities/markets/userBets.ts
Diogomartf: Yeah, makes sense, I will update it to `UserBets`.

app/(main)/my-bets/page.tsx
```diff
@@ -15,41 +15,28 @@ import { BetsListPanel, BetsListTab } from '@/app/components';
...
-        const marketCondition = new MarketCondition(
-          userPosition.fpmm,
-          position.condition
-        );
-        const canRedeem = marketCondition.canRedeem(outcomeIndex, userPosition.balance);
-        return canRedeem;
-      }) ?? [],
-    [userPositionsComplete]
+    () => userBetsManager.getUnredeemedBets(),
+    [userBetsManager]
```
ElRodrigote, line 39: This looks so much more clean and elegant, great work!!

ElRodrigote (approved):

PR #135, opened 2024-10-28
[FIX][PGIO-112] - TransactionModal component accessibility

ElRodrigote:
## Description
* Implements a hidden `DialogTitle` and `DialogDescription` for the Transaction Modal a11y
* ~Temporary hides a11y warnings until we get a `DialogDescription` component exported from `@swapr/ui`, and we can create and hide an a11y compliant dialog description~


berteotti (approved):

app/components/TransactionModal.tsx
```diff
@@ -34,7 +37,14 @@ export const TransactionModal = ({
   return (
     <Dialog open={isModalOpen(ModalId.WAITING_TRANSACTION)} onOpenChange={close}>
       <DialogContent>
-        <DialogHeader />
+        <DialogHeader>
+          <VisuallyHidden asChild>
+            <>
+              <DialogTitle>Transaction Modal</DialogTitle>
+              <DialogDescription />
```
Diogomartf, line 44: do we need the `DialogDescription`?

Diogomartf (approved):

app/components/TransactionModal.tsx
ElRodrigote: Yes, both Title and Description are required by Content:
* The Title missing triggers an error severity console message
* The Description missing triggers a warning severity console message

Placing both components inside the Content DOM tree removes the error and the warning.
`VisuallyHidden` can only have 1 children, that's why I used a Fragment to wrap both the Title and the Description

app/components/TransactionModal.tsx
```diff
@@ -34,7 +37,14 @@ export const TransactionModal = ({
   return (
     <Dialog open={isModalOpen(ModalId.WAITING_TRANSACTION)} onOpenChange={close}>
       <DialogContent>
-        <DialogHeader />
+        <DialogHeader>
+          <VisuallyHidden asChild>
+            <>
```
berteotti, line 42: Do we need the Fragment?

ElRodrigote: Yes :c

`VisuallyHidden` can only have 1 children, that's why I used a Fragment to wrap both the Title and the Description

PR #141, opened 2024-11-19
adds bets icon to "My bets" button

Diogomartf:
Adds a nice touch to the "my bets" button

image
image


ElRodrigote (approved):

app/components/Navbar.tsx
```diff
@@ -61,10 +62,16 @@ export const Navbar = () => {
...
+                <Image
+                  src="assets/icons/bets.svg"
+                  width={16}
+                  height={16}
+                  alt="my bets icon"
+                />
+                <p> My bets</p>
               </Button>
               {hasUnredeemedBets && (
                 <div className="absolute -right-2 -top-1 flex size-5 items-center justify-center rounded-100 border-2 border-surface-surface-bg bg-surface-success-main p-1 text-text-white">
```
berteotti, line 77: Can we fix the text color has well?

Diogomartf: what's wrong with the text color?

berteotti: low visibility on dark mode

berteotti: something like text-text-neutral-alt-white

Diogomartf: I think you meant the icon color, as the text is the same as it was (an the others), I'll take a look. 
The idea was to have a lit bit less contrast.

berteotti: I'm suggesting to change the unredeem bets bubble text color, it works on light mode but in dark mode is not easy to read. The green brightness merges with the text white color, creating less contrast with the background, so it becomes difficult to read in dark mode.

PR #145, opened 2024-11-22
feat: add and remove liquidity from market

berteotti:
Enables users to add and remove liquidity from markets.

image

### DIFF
--- a/app/(main)/markets/Liquidity.tsx
+++ b/app/(main)/markets/Liquidity.tsx
@@ -1,0 +1,476 @@
+'use client';
+
+import {
+  TokenLogo,
+  SwapInput,
+  ConnectButton,
+  Spinner,
+  succesApprovalTxToast,
+  waitingTxToast,
+} from '@/app/components';
+import { useQuery } from '@tanstack/react-query';
+import { Query } from '@/queries/omen';
+import {
+  Button,
+  Dialog,
+  DialogBody,
+  DialogClose,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+  Icon,
+  ToggleGroup,
+  ToggleGroupOption,
+  VisuallyHidden,
+  errorToast,
+} from '@swapr/ui';
+import { Abi, Address, erc20Abi, formatEther, parseEther } from 'viem';
+import { Market, Token } from '@/entities';
+import { useState } from 'react';
+import { ModalId, useModal, useTx } from '@/context';
+import { formatEtherWithFixedDecimals } from '@/utils';
+import {
+  calcAddFundingOutcomeTokensReturned,
+  calcRemoveFundingSendAmounts,
+  calcTokensToReceiveAfterAddFunding,
+} from '@/utils/liquidity';
+import { useAccount, useConfig, useReadContract, useWriteContract } from 'wagmi';
+import { MarketABI } from '@/abi';
+import {
+  useReadAllowance,
+  useReadBalanceOf,
+  useReadToken,
+} from '@/hooks/contracts/erc20';
+import {
+  waitForTransactionReceipt,
+  WaitForTransactionReceiptErrorType,
+  WriteContractErrorType,
+} from 'wagmi/actions';
+import { ChainId } from '@/constants';
+import { cx } from 'class-variance-authority';
+import { useUnsupportedNetwork } from '@/hooks';
+
+enum LiquidityOperation {
+  ADD = 'add',
+  REMOVE = 'remove',
+}
+
+export const Liquidity = ({ id }: { id: Address }) => {
+  const [amount, setAmount] = useState('');
+  const [liquidityOperation, setLiquidityOperation] = useState(LiquidityOperation.ADD);
+  const [isApproving, setIsApproving] = useState(false);
+  const amountWei = parseFloat(amount) > 0 ? parseEther(amount) : BigInt(0);
+  const { isModalOpen, closeModal, openModal } = useModal();
+  const { address, isDisconnected } = useAccount();
+  const unsupportedNetwork = useUnsupportedNetwork();
+  const { data } = useQuery<Pick<Query, 'fixedProductMarketMaker'>>({
+    queryKey: ['getMarket', id],
+  });
+  const { submitTx } = useTx();
+  const config = useConfig();
+
+  const { data: totalSupply } = useReadContract({
+    abi: MarketABI,
+    address: id,
+    functionName: 'totalSupply',
+  });
+
+  const { data: fundingBalance, refetch: refetchFundingBalance } = useReadBalanceOf({
+    tokenAddress: id,
+    address,
+  });
+
+  const tokenAddress = data?.fixedProductMarketMaker?.collateralToken;
+
+  const { symbol, decimals, name } = useReadToken({
+    tokenAddress,
+  });
+
+  const { data: balance, refetch: refetchCollateralBalance } = useReadBalanceOf({
+    tokenAddress,
+    address,
+  });
+  const { data: allowance, refetch: refetchCollateralAllowence } = useReadAllowance({
+    address,
+    tokenAddress,
+    spenderAddress: id,
+  });
+
+  const { writeContractAsync } = useWriteContract();
+
+  if (
+    !data?.fixedProductMarketMaker ||
+    totalSupply === undefined ||
+    !symbol ||
+    !decimals ||
+    !name
+  )
+    return;
+
+  const fixedProductMarketMaker = data.fixedProductMarketMaker;
+  const marketModel = new Market(fixedProductMarketMaker);
+
+  const sharesToReceive =
+    parseFloat(amount) > 0
+      ? formatEtherWithFixedDecimals(
+          calcTokensToReceiveAfterAddFunding(
+            amountWei,
+            fixedProductMarketMaker.outcomeTokenAmounts.map(amount => BigInt(amount)),
+            totalSupply as bigint
+          ),
+          2
+        )
+      : null;
+
+  const outcomeTokensToReceive =
+    parseFloat(amount) > 0
+      ? calcAddFundingOutcomeTokensReturned(
+          amountWei,
+          fixedProductMarketMaker.outcomeTokenAmounts.map(amount => BigInt(amount)),
+          totalSupply as bigint
+        )
+      : [];
+
+  const collateralToReceive =
+    parseFloat(amount) > 0
+      ? calcRemoveFundingSendAmounts(
+          amountWei,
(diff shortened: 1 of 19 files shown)


Diogomartf:
We should show that you provided liquidity on a market, maybe adding LP somewhere (maybe next to liquidity).

Also on the liquidity tab we show how much you provided (pool tokens balance), and maybe details like when (date), etc.


I'm an LP of this market but the only way to know is go to remove tab and check balance.
image

app/components/TransactionModal.tsx
```diff
@@ -38,9 +38,10 @@ export const TransactionModal = ({
     <Dialog open={isModalOpen(ModalId.WAITING_TRANSACTION)} onOpenChange={close}>
       <DialogContent>
         <DialogHeader>
+          {/* This looks akward */}
```
Diogomartf, line 41:
```suggestion
```

app/components/TransactionModal.tsx
```diff
@@ -38,9 +38,10 @@ export const TransactionModal = ({
...
           <VisuallyHidden asChild>
             <>
-              <DialogTitle>Transaction Modal</DialogTitle>
+              <DialogTitle></DialogTitle>
```
Diogomartf, line 44: why?

Diogomartf: I think we need this to describe the aria label

app/components/toasts.tsx
Diogomartf, line 1: Nice!

app/components/toasts.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <Spinner className="h-5 w-5 shrink-0 animate-spin" />
+        <div className="font-normal">
+          Wating for transaction confirmation{' '}
+          <a
+            href={`${GNOSIS_SCAN_URL}/tx/${txHash}`}
+            target="_blank"
+            className="inline-block underline"
+          >
+            {shortenAddress(txHash)}
+          </a>
```
Diogomartf, line 41: maybe we could extract this as a function here, as it's the same as above

app/components/ui/SwapInput.tsx
```diff
@@ -37,7 +37,7 @@ export const SwapInput = ({
           className="w-full overflow-hidden overscroll-none bg-transparent text-3xl caret-text-primary-main outline-none placeholder:text-text-disabled"
           {...props}
         />
-        {selectedToken instanceof Token ? (
+        {selectedToken === null ? null : selectedToken instanceof Token ? (
```
Diogomartf, line 40: this looks a bit odd :/

queries/omen/markets.ts
```diff
@@ -51,6 +50,7 @@ const getMarketQuery = gql`
       answerFinalizedTimestamp
       scaledLiquidityParameter
       scaledLiquidityMeasure
+      liquidityMeasure
```
Diogomartf, line 53: what is this field?

app/(main)/markets/MarketDetails.tsx
```diff
@@ -164,6 +166,16 @@ export const MarketDetails = ({ id }: MarketDetailsProps) => {
                 <Info fixedProductMarketMaker={fixedProductMarketMaker} />
               </div>
             )}
+            {tab === Tabs.INFO && (
+              <div className="mx-4 my-2 flex flex-col divide-y divide-outline-low-em">
+                <Info fixedProductMarketMaker={fixedProductMarketMaker} />
+              </div>
+            )}
```
Diogomartf, line 173: this is repeated

```suggestion

```

app/components/TransactionModal.tsx
berteotti: 1. This is not being properly hidden, show "Transaction Modal" text appears.
2. If you hide the tittle, the modal layout breaks (close button goes to the left). Maybe there is another solution or we just need to improve the lib component.

queries/omen/markets.ts
berteotti: This is liquidity but with not formatted to decimal number. Will delete it since it's not being used anymore.

app/components/ui/SwapInput.tsx
ElRodrigote: Yeah agree, maybe we could do something like abstracting this into a new component in this same file, and do an early return if no `selectedToken` is provided

```tsx
interface ComponentProps {
  selectedToken: Token | Outcome | null;
  onTokenClick?: (outcome?: Outcome) => void;
  tokenList: Array<Outcome> | null;
}

const Component = ({ selectedToken, tokenList, onTokenClick }: ComponentProps) => {
  if (!selectedToken) return null;

  return selectedToken instanceof Token ? (
    <Button className="flex-shrink-0" variant="pastel">
     /* ... */
    </Button>
  ) : (
    <Popover>
      /* ... */
    </Popover>
  );
};

```

Then inside the `SwapInput` we do: 
```tsx
  return (
    <div className="space-y-1 rounded-16 bg-surface-surface-1 p-4">
      <div className="text-xs font-semibold uppercase text-text-low-em">{title}</div>
      <div className="flex w-full space-x-2">
        <input
          ...
        />
        <Component
          selectedToken={selectedToken}
          tokenList={tokenList}
          onTokenClick={onTokenClick}
        />
      </div>
      {children}
    </div>
  );
```

utils/liquidity.ts
```diff
@@ -0,0 +1,199 @@
...
+  const allEqual = initialOdds.every(initialOdd => initialOdd === initialOdds[0]);
+  if (allEqual) {
+    return [];
+  }
+  const initialOddsBig = initialOdds.map(x => new Big(x));
+  const product = initialOddsBig.reduce((a, b) => a.mul(b));
+
+  const distributionHint = initialOddsBig
+    .map(o => product.div(o))
+    .map(x => x.mul(1000000).round())
```
ElRodrigote, line 95: What's that `1000000` ? The amount of decimals?

utils/liquidity.ts
```diff
@@ -0,0 +1,199 @@
...
+    poolShareSupply
+  );
+  return sendAmounts.reduce((min, amount) => (amount < min ? amount : min));
+};
+
+/**
+ * Compute the numbers of outcome tokens that will be added to the market maker after adding `addedFunds` of collateral.
+ * @param addedFunds - the amount of collateral being added to the market maker as liquidity
+ * @param poolBalances - the market maker's balances of outcome tokens
+ */
```
ElRodrigote, line 64:
```suggestion
/**
 * Compute the numbers of outcome tokens that will be added to the market maker after adding `addedFunds` of collateral.
 * @param addedFunds - the amount of collateral being added to the market maker as liquidity
 * @param poolBalances - the market maker's balances of outcome tokens
 * @returns an array of BigInt values representing the deposited amounts
 * @throws an error if any pool balance is zero
 */
```

utils/liquidity.ts
```diff
@@ -0,0 +1,199 @@
...
+  const poolWeight = poolBalances.reduce((a, b) => (a > b ? a : b));
+
+  const depositAmounts = poolBalances.map(h => (addedFunds * h) / poolWeight);
+
+  return depositAmounts;
+};
+/**
+ * Computes the distribution hint that should be used for setting the initial odds to `initialOdds`
+ * @param initialOdds - an array of numbers proportional to the initial estimate of the probability of each outcome
+ */
```
ElRodrigote, line 84:
```suggestion
/**
 * Computes the distribution hint that should be used for setting the initial odds to `initialOdds`
 * @param initialOdds - an array of numbers proportional to the initial estimate of the probability of each outcome
 * @returns an array of BigInt values representing the distribution hint
 */
```

app/(main)/markets/Liquidity.tsx
```diff
@@ -0,0 +1,484 @@
...
+      ) : (
+        <Button
+          width="full"
+          variant="pastel"
+          size="lg"
+          onClick={() => openModal(ModalId.CONFIRM_LIQUIDITY)}
+        >
+          {activeLiquidityOperationState.actionTitle}
+        </Button>
+      )}
```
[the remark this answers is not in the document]
berteotti: I think we can abstract this logic into a new component. We can do it in a new task.

utils/liquidity.ts
berteotti: Looks like it. This code isn't being used but might be useful for market creation. So we need to review this code again when building market creation feature.

app/(main)/markets/Liquidity.tsx
```diff
@@ -0,0 +1,483 @@
...
+    !decimals ||
+    !name
+  )
+    return;
+
+  const fixedProductMarketMaker = data.fixedProductMarketMaker;
+  const marketModel = new Market(fixedProductMarketMaker);
+
+  const unsupportedNetwork =
+    !!chainId && !supportedChains.some(supportedChain => supportedChain.id === chainId);
```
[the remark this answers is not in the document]
berteotti: Created a new hook for this. Thanks for highlighting this!

app/components/ui/SwapInput.tsx
berteotti: Found a simple solution for it. Let me know if it doesn't work.

utils/liquidity.ts
Diogomartf: maybe delete it till then ? OR as least leave a comment there?

Also as this code is being copied from elsewhere, can we add a comment with a link to that on the top of the file?

app/components/TransactionModal.tsx
```diff
@@ -38,12 +37,8 @@ export const TransactionModal = ({
...
       <DialogContent>
         <DialogHeader>
-          <VisuallyHidden asChild>
-            <>
-              <DialogTitle>Transaction Modal</DialogTitle>
-              <DialogDescription />
-            </>
-          </VisuallyHidden>
+          <DialogTitle></DialogTitle>
+          <DialogDescription></DialogDescription>
```
Diogomartf, line 41: why?

the idea of having `DialogTitle` is not just for not having the warning but also to provide label for aria description. In this case we lose that, right?

app/components/TransactionModal.tsx
Diogomartf: that's odd, cause the whole purpose we add it was for having hidden element to do descriptions, maybe it's because of `asChild` props?  (if you could try)

We should fix this (in another PR) not sure if we should change it in this PR tho.

Diogomartf:
image

About showing the you're a LP of the market. I think we should add something in this PR.

Also we show the pool shares (if user have some) when on liquidity tab

app/components/TransactionModal.tsx
berteotti: Found a solution to this. But I think we can simplify on the UI lib. Thanks for pointing this out again

ElRodrigote (approved):

Diogomartf (approved):
Great job Leo, this one was big task and you nailed it really well. 👏

PR #146, opened 2024-12-12
feat(PGIO-139): Update the OutcomeBar for closed markets

ElRodrigote:
## Fixes: [PGIO-139](https://linear.app/swaprhq/issue/PGIO-139/users-should-be-able-to-see-what-resolution-answer-at-the-top-of-tabs)

## Description
* Abstracts logic into a reusable `OutcomeTag` component, adding borders if needed
* Implements the new component inside the Bets tab and at the top of the Market Details outcome

### Light mode
Screenshot 2024-12-12 at 1 02 24 PM
Screenshot 2024-12-12 at 1 22 53 PM
Screenshot 2024-12-12 at 1 20 46 PM

### Dark mode
Screenshot 2024-12-12 at 1 21 19 PM
Screenshot 2024-12-12 at 1 21 24 PM
Screenshot 2024-12-12 at 1 21 29 PM


app/(main)/markets/MarketDetails.tsx
```diff
@@ -81,8 +83,18 @@ export const MarketDetails = ({ id }: MarketDetailsProps) => {
...
                 </a>
                 {marketModel.isClosed ? (
-                  <Tag className="w-fit capitalize" size="sm" colorScheme="quaternary">
-                    Market Closed
+                  <Tag
+                    className="w-fit gap-1 capitalize"
+                    colorScheme="quaternary"
+                    size="sm"
+                  >
+                    <p>Market Closed: Outcome</p>
```
Diogomartf, line 91: let's simplify, it does not need "outcome" text we can use a html title with "the outcome for this market is NO" 

```suggestion
                    <p>Market Closed</p>
```

app/components/OutcomeBar.tsx
```diff
@@ -95,32 +99,59 @@ export const OutcomeBar = ({ market }: OutcomeBarProps) => {
...
       </div>
 
       <div className="flex h-4 justify-between text-sm font-semibold">
         {hasOutcomePercentages && (
           <>
-            <p className="w-full uppercase text-text-success-main">{`${outcome0.symbol} ${outcome0.percentage || '-'}%`}</p>
+            <p className="w-full uppercase text-text-success-main">{`${outcome0.symbol} ${outcome0percentage || '-'}%`}</p>
             <p className="w-full text-right uppercase text-text-danger-main">
-              {`${outcome1.symbol} ${outcome1.percentage || '-'}%`}
+              {`${outcome1.symbol} ${outcome1percentage || '-'}%`}
```
Diogomartf, line 154: **Suggestion**
Add this logic in to function

app/components/OutcomeBar.tsx
```diff
@@ -95,32 +99,63 @@ export const OutcomeBar = ({ market }: OutcomeBarProps) => {
...
-            width: outcome1?.percentage ? `${outcome1.percentage}%` : '50%',
-          }}
-        />
+        {showOutcome0 && (
+          <div
+            className={cx(
+              'flex h-3 items-center rounded-s-8 bg-surface-success-accent-2 px-2',
+              {
+                'rounded-e-8': !showOutcome1,
+              }
```
Diogomartf, line 126: I don't think we need the braces, might be wrong tho, would be cleaner if not.
```suggestion
                'rounded-e-8': !showOutcome1,
```

app/components/OutcomeBar.tsx
```diff
@@ -95,32 +99,63 @@ export const OutcomeBar = ({ market }: OutcomeBarProps) => {
...
+          />
+        )}
+
+        {showOutcome1 && (
+          <div
+            className={cx(
+              'flex h-3 items-center rounded-e-8 bg-surface-danger-accent-2 px-2',
+              {
+                'rounded-s-8': !showOutcome0,
+              }
```
Diogomartf, line 140: same here, if no braces work, let's remove them.

app/components/OutcomeBar.tsx
```diff
@@ -95,32 +99,63 @@ export const OutcomeBar = ({ market }: OutcomeBarProps) => {
...
-            <p className="w-full uppercase text-text-success-main">{`${outcome0.symbol} ${outcome0.percentage || '-'}%`}</p>
-            <p className="w-full text-right uppercase text-text-danger-main">
-              {`${outcome1.symbol} ${outcome1.percentage || '-'}%`}
+            <p
+              className={`w-full uppercase ${showOutcome0 ? 'text-text-success-main' : 'text-text-low-em'}`}
+            >{`${outcome0.symbol} ${outcome0percentage || '-'}%`}</p>
+            <p
+              className={`w-full text-right uppercase ${showOutcome1 ? 'text-text-danger-main' : 'text-text-low-em'}`}
+            >
+              {`${outcome1.symbol} ${outcome1percentage || '-'}%`}
```
Diogomartf, line 158: Suggestion, we could extract this logic into a function

Diogomartf (approved):

app/components/OutcomeBar.tsx
```diff
@@ -95,32 +99,63 @@ export const OutcomeBar = ({ market }: OutcomeBarProps) => {
...
+
+    return [outcome0?.percentage, outcome1?.percentage];
+  };
+
+  const outcomesPercentages = getOutcomesPercentages();
+  const outcome0percentage = outcomesPercentages[0];
+  const outcome1percentage = outcomesPercentages[1];
+  const hasOutcomePercentages = outcome0percentage && outcome1percentage;
+  const showOutcome0 = !!Number(outcome0percentage) && isAnswerFinal;
+  const showOutcome1 = !!Number(outcome1percentage) && isAnswerFinal;
```
berteotti, line 115: We should call it `isWinnerOutcome0` and maybe check `winnerOutcome.index` instead of `outcome0percentage`.

```suggestion
  const isWinnerOutcome0 = winnerOutcome.index === 0 && isAnswerFinal;
  const isWinnerOutcome1 = winnerOutcome.index === 1 && isAnswerFinal;
```

berteotti: I got it now why it's called show. Maybe we can keep the show, but I think it's cleaner to call the variable what it is instead of what it does.

app/components/OutcomeBar.tsx
ElRodrigote: `cx` is basically a `clsx` extension, in the CVA docs they suggest to check `clsx` usage:
https://github.com/lukeed/clsx?tab=readme-ov-file#usage

I think there's a way to use a shortcircuit instead the braces.

berteotti (approved):

app/components/OutcomeBar.tsx
```diff
@@ -39,6 +40,10 @@ type OutcomeTokenMarginalPricesResponse = {
 
 export const OutcomeBar = ({ market }: OutcomeBarProps) => {
   const { id } = market;
+  const marketModel = new Market(market);
+  const winnerOutcome = marketModel.getWinnerOutcome();
+  const isClosed = marketModel.isClosed;
+  const isClosedAndInvalid = isClosed && !winnerOutcome;
```
berteotti, line 46: Why not use `marketModel.isAnswerInvalid` here instead? We are showing grey on finalising and pending markets, although they don't have an answer and are closed, they are not invalid.

berteotti: Here is a visual example

image

PR #147, opened 2024-12-16
fix(PGIO-139): use isAnswerInvalid for checking if the market is invalid

ElRodrigote:
## Fixes: [PGIO-139](https://linear.app/swaprhq/issue/PGIO-139/users-should-be-able-to-see-what-resolution-answer-at-the-top-of-tabs)

## Description
* Updates the `OutcomeBar` component logic to use `marketModel.isAnswerInvalid` (from the subgraph) instead of negating the result from `marketModel.getWinnerOutcome()`


berteotti:
Let's have the text also in red/green instead of gray here

image

berteotti (approved):

Diogomartf (approved):

PR #148, opened 2024-12-17
feat(PGIO-135): abstract action button logic

ElRodrigote:
## Fixes: [PGIO-135](https://linear.app/swaprhq/issue/PGIO-135/abstract-action-button-logic)

## Description
* Abstracts the Disconnected and Invalid Network logic into a wrapper component
* Implement the new button in the Liquidity and Swapbox components to re-use logic


berteotti:
I think the button is doing to much logic. Let's abstract only  Connect and unsupported network logic in the begging. After that we can check again if it makes sense to abstract more or if it's enough.

app/components/Swapbox.tsx
```diff
@@ -362,30 +360,30 @@ export const Swapbox = ({ fixedProductMarketMaker }: SwapboxProps) => {
...
-            </Button>
-          ) : !tokenAmountIn || +tokenAmountIn === 0 ? (
-            <Button width="full" variant="pastel" size="lg" disabled>
-              Enter amount
-            </Button>
-          ) : currentState.isLoading ? (
+          {currentState?.isLoading ? (
             <Button width="full" variant="pastel" size="lg" disabled>
               Fetching price
             </Button>
```
berteotti, line 366: This can also be inside the button wrapper, right?

berteotti: I see it's duplicated. We can remove this one

app/components/ExecuteTxButtonWrapper.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import { PropsWithChildren } from 'react';
+
+import { useAccount } from 'wagmi';
+
+import { ConnectButton } from '@/app/components';
+import { useUnsupportedNetwork } from '@/hooks';
+import { Button } from '@swapr/ui';
+
+export const ExecuteTxButtonWrapper = ({ children }: PropsWithChildren) => {
```
berteotti, line 11: Maybe I would call it `AccountStateButton` for now. What do you think?

app/components/Swapbox.tsx
```diff
@@ -362,31 +360,25 @@ export const Swapbox = ({ fixedProductMarketMaker }: SwapboxProps) => {
...
+                Fetching price
+              </Button>
+            ) : +tokenAmountIn > +formatEther(currentState.balance) ? (
+              <Button width="full" variant="pastel" size="lg" disabled>
+                Insufficient {currentState.inToken.symbol} balance
+              </Button>
+            ) : (
+              <Button width="full" variant="pastel" size="lg" onClick={openBetModal}>
+                {currentState.buttonText}
+              </Button>
```
berteotti, line 379: We can also abstract the button style props and receive `onClick` and `disabled` props. What do you think?

ElRodrigote: I think the whole structure can be abstracted, that's why I did it that way because what changes are the balances, the amounts, the "final" button label, and the final button `onClick` function. The only actionable buttons are the "final" and the Connect buttons.

For this specific suggestion, if we go this way I'd abstract the `onClick` and `variant` props + the `children` (the button label) and move this Button inside the AccountStateButton. WDYT?

berteotti: I wouldn't abstract everything. For example, current logic inside the `children` would be the same but only without the button element. I would let the developer define things like it was a normal button. This way it would be more flexible.

I'm suggesting this because it's the only thing in common with all the Button elements. Style, disable and onClick props are used "everywhere" (disabled and onClick are omitted and some buttons). Otherwise, the logic outside this is different for each use case.

Diogomartf:
the preview is breaking

image

**How to reproduce**
- open landing
- enter a market
- try to use swap input and it breaks

app/components/AccountStateButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import { PropsWithChildren } from 'react';
+
+import { useAccount } from 'wagmi';
+
+import { ConnectButton } from '@/app/components';
+import { useUnsupportedNetwork } from '@/hooks';
+import { Button } from '@swapr/ui';
+
+export const AccountStateButton = ({ children }: PropsWithChildren) => {
```
Diogomartf, line 11: I'm not a fan of adding state to the name, `AccountButton`, sounds better and implies relationship with account. 

And also not sure if is account is the right name, as I read it, I associate more with the Navbar connect button, that's for me is account. Ability to connect and showing address/balance when connected. 

image

So we probably need to find a better name. 

The case for this component is that we want to make it easier for the user to login if he is not connected, or change to the right network if the user is on different one.

these methods: `isDisconnected`, `unsupportedNetwork`

So we add that capability to extend the button.
I'll give some suggestion, not sure if is good but I think they are better describing:

`ButtonWithConnect` (the idea here it to imply that that's a button has connect capabilities)

utils/buttons.ts
```diff
@@ -0,0 +1,10 @@
+export const getButtonLabel = (
+  amount: number,
+  defaultLabel: string,
+  isEnoughBalance: boolean,
+  tokenSymbolText: string
+) => {
+  if (!amount) return 'Enter amount';
+  else if (!isEnoughBalance) return `Insufficient ${tokenSymbolText} balance`;
+  else return defaultLabel;
+};
```
Diogomartf, line 10: why not this in button component?  I don't feel `utils` is the right place.

app/components/Swapbox.tsx
```diff
@@ -362,31 +370,20 @@ export const Swapbox = ({ fixedProductMarketMaker }: SwapboxProps) => {
...
+            size="lg"
+            variant="pastel"
+            width="full"
+          >
+            {getButtonLabel(
+              +tokenAmountIn,
+              defaultButtonLabel,
+              isEnoughBalance,
+              tokenSymbolText
+            )}
```
berteotti, line 385: I think it's better to have the condition logic here in the component.
It can be inside a function, but should be inside the component.

Let's not abstract this logic for now.

app/components/Swapbox.tsx
```diff
@@ -362,31 +370,20 @@ export const Swapbox = ({ fixedProductMarketMaker }: SwapboxProps) => {
...
-            <Button width="full" variant="pastel" size="lg" disabled>
-              Insufficient {currentState.inToken.symbol} balance
-            </Button>
-          ) : (
-            <Button width="full" variant="pastel" size="lg" onClick={openBetModal}>
-              {currentState.buttonText}
-            </Button>
-          )}
+          <AccountStateButton
+            disabled={isButtonDisabled}
```
berteotti, line 374: I think this logic is missing

```
currentState.isLoading ? (
            <Button width="full" variant="pastel" size="lg" disabled>
              Fetching price
            </Button>
          )
```

app/(main)/markets/Liquidity.tsx
```diff
@@ -257,6 +255,19 @@ export const Liquidity = ({ id }: { id: Address }) => {
...
+
+  const getButtonLabel = () => {
+    const tokenSymbolText =
+      liquidityOperation === LiquidityOperation.ADD
+        ? liquidityOperationState[liquidityOperation].inToken.symbol ?? 'pool tokens'
+        : 'pool tokens';
+
+    if (!amount) return 'Enter amount';
+    else if (!isEnoughBalance) return `Insufficient ${tokenSymbolText} balance`;
+    else return activeLiquidityOperationState.actionTitle;
```
berteotti, line 269: since we have return inside each if, we can remove the else

berteotti: `hasEnoughBalance` sounds better IMO

```suggestion
    else if (!hasEnoughBalance) return `Insufficient ${tokenSymbolText} balance`;
    else return activeLiquidityOperationState.actionTitle;
```

app/components/Swapbox.tsx
```diff
@@ -362,30 +371,20 @@ export const Swapbox = ({ fixedProductMarketMaker }: SwapboxProps) => {
...
-            </Button>
-          ) : !tokenAmountIn || +tokenAmountIn === 0 ? (
-            <Button width="full" variant="pastel" size="lg" disabled>
-              Enter amount
-            </Button>
-          ) : currentState.isLoading ? (
+          {currentState.isLoading ? (
             <Button width="full" variant="pastel" size="lg" disabled>
               Fetching price
             </Button>
```
berteotti, line 377: We isn't this inside the `getButtonLabel` and `disable` logic? I think we can add it there too

app/components/Swapbox.tsx
```diff
@@ -272,6 +270,17 @@ export const Swapbox = ({ fixedProductMarketMaker }: SwapboxProps) => {
...
 
+  const isEnoughBalance =
+    +tokenAmountIn <= +formatEther(currentState.balance ?? BigInt(0));
+  const isButtonDisabled = !tokenAmountIn || !isEnoughBalance;
+
+  const getButtonLabel = () => {
+    if (!tokenAmountIn) return 'Enter amount';
+    else if (!isEnoughBalance)
+      return `Insufficient ${currentState.inToken.symbol ?? 'pool tokens'} balance`;
+    else return currentState.buttonText;
```
berteotti, line 281: Some comment about the else and the `hasEnoughBalance` here

app/(main)/markets/Liquidity.tsx
```diff
@@ -313,36 +324,16 @@ export const Liquidity = ({ id }: { id: Address }) => {
...
-        >
-          {activeLiquidityOperationState.actionTitle}
-        </Button>
-      )}
+      <TxButton
+        disabled={isButtonDisabled}
+        onClick={() => openModal(ModalId.CONFIRM_LIQUIDITY)}
+        size="lg"
+        variant="pastel"
+        width="full"
```
berteotti, line 332: We can move this into the component / abstract it, right?

app/components/Swapbox.tsx
```diff
@@ -362,30 +371,20 @@ export const Swapbox = ({ fixedProductMarketMaker }: SwapboxProps) => {
...
           ) : (
-            <Button width="full" variant="pastel" size="lg" onClick={openBetModal}>
-              {currentState.buttonText}
-            </Button>
+            <TxButton
+              disabled={isButtonDisabled}
+              onClick={openBetModal}
+              size="lg"
+              variant="pastel"
+              width="full"
```
berteotti, line 384: Also we can remove this if we abstract

app/components/Swapbox.tsx
Diogomartf, line 281: I think it's cleaner like

```suggestion
    if (!tokenAmountIn) return 'Enter amount';
    
    if (!isEnoughBalance)
      return `Insufficient ${currentState.inToken.symbol ?? 'pool tokens'} balance`;
    return currentState.buttonText;
```

berteotti (approved):
Looking awesome!

PR #149, opened 2024-12-23
feat: add market creation

berteotti:
Adds support for binary market creation. 

Create market route - https://feat-pgio-105-support-for-ma.presagio.pages.dev/create-market

Please create markets with "test" category.

Screenshot 2024-12-23 at 11 43 31


app/(main)/create-market/page.tsx
```diff
@@ -0,0 +1,671 @@
...
+    else if (outcomes.findIndex(outcome => !outcome.name) !== -1)
+      setError('Enter outcome names');
+    else if (!category) setError('Enter category');
+    else if (!amount || +amount === 0) setError('Enter amount');
+    else if (dateTimeError) setError('Enter valid closing date');
+    else if (percentagesError) setError('Enter valid outcomes');
+    else setError(null);
+  }, [amount, category, dateTimeError, outcomes, percentagesError, question]);
+
+  const maxBalance = useCallback(() => {
```
ElRodrigote, line 103: Is there a significant benefit in wrapping all these state setters with the `useCallback` hook? I just tried to profile a bit this component page with and without these hooks, and by just removing the usage of `useCallback` we gain +8% perfo in Lighthouse

With the hooks in
Screenshot 2024-12-24 at 10 00 48 AM

Without the hooks
Screenshot 2024-12-24 at 11 06 01 AM

ElRodrigote:
Some things I'd like to see on the market creation flow (not necessarily all of them, but any combination of them):
* The market creation modal should not be dismissed when we click on the modal backdrop: if the user misclicks on the backdrop they'll lose the context on the market creation flow/steps
* A small stepper or something like that lets the user know how many signature/steps are done already and how many are missing. Maybe a 1/2/3/4 (step number) with a small description label beneath the step number, and then the number gets replaced with a spinner if we're waiting for its response. Something similar to what bridges do when using them:
ask question > prepare condition > approve token > create market (create2FixedProductMarketMaker)
* A toast/snackbar appearing when the whole process is done or to display an error if something fails (for now, as there's no error handling, the flow gets stuck if the user cancels a TX)
* On market creation success, maybe a redirection to the specific market details or to the markets list filtered by the newly created market category so you can see the result of your action would be awesome.

app/(main)/create-market/page.tsx
```diff
@@ -0,0 +1,671 @@
...
+  percentage: number;
+};
+
+const defaultInputOutcomes: InputOutcome[] = [
+  { name: '', percentage: 50 },
+  { name: '', percentage: 50 },
+];
+const tokenList = [WXDAI, SDAI];
+
+const LP_FEE = parseEther('0.02');
```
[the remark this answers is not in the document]
berteotti: Currently this is fixed, but should be updatable by the user. Maybe we can add this in the future.

ElRodrigote (approved):

Diogomartf:
can we change the route for `/markets/create`?  This route is feels much better.

Diogomartf:
We can change title to only "Preview" it's simpler.

image

app/(main)/create-market/page.tsx
Diogomartf, line 1: lets change the folder structure to be `/markets/create/page.tsx` this will make the url without the slash, which is cleaner: /markets/create

In our case we already have the markets folder, we just need a create folder inside with page file.

app/components/ui/SwapInput.tsx
```diff
@@ -9,10 +9,11 @@ import { TokenLogo } from '../TokenLogo';
 interface SwapInputProps extends InputHTMLAttributes<HTMLInputElement> {
   title: string;
   selectedToken: Token | Outcome | null;
-  onTokenClick?: (outcome?: Outcome) => void;
-  tokenList: Array<Outcome> | null;
+  onTokenClick?: (asset?: Token | Outcome) => void;
+  tokenList: Array<Token | Outcome> | null;
 }
 
+//TODO create 2 different instances, 1 for token and other for outcomes
```
Diogomartf, line 16: should we handle this comment on this PR?

hooks/contracts/kleros.ts
```diff
@@ -0,0 +1 @@
+export const KLEROS_ARBITRATOR_ADDRESS = '0xe40DD83a262da3f56976038F1554Fe541Fa75ecd';
```
Diogomartf, line 1: can't we use constants file for this? I would not create a file just for this constant.

Diogomartf: Also inside hooks, doesn't make sense for me.

hooks/contracts/realityeth.ts
```diff
@@ -0,0 +1,4 @@
+export const REALITY_ETH_CONTRACT_ADDRESS = '0x79e32aE03fb27B07C89c0c568F80287C01ca2E57';
+export const ORACLE_ADDRESS = '0xAB16D643bA051C11962DA645f74632d3130c81E2';
+export const REALITY_ETH_TIMEOUT = 86400;
+export const SINGLE_SELECT_TEMPLATE_ID = 2;
```
Diogomartf, line 4: why have inside hooks?

Why all of the constants together in the same file ?

Diogomartf:
> Some things I'd like to see on the market creation flow (not necessarily all of them, but any combination of them):
> 
> * The market creation modal should not be dismissed when we click on the modal backdrop: if the user misclicks on the backdrop they'll lose the context on the market creation flow/steps
> * A small stepper or something like that lets the user know how many signature/steps are done already and how many are missing. Maybe a 1/2/3/4 (step number) with a small description label beneath the step number, and then the number gets replaced with a spinner if we're waiting for its response. Something similar to what bridges do when using them:
>   ask question > prepare condition > approve token > create market (create2FixedProductMarketMaker)
> * A toast/snackbar appearing when the whole process is done or to display an error if something fails (for now, as there's no error handling, the flow gets stuck if the user cancels a TX)
> * On market creation success, maybe a redirection to the specific market details or to the markets list filtered by the newly created market category so you can see the result of your action would be awesome.

Agree on this, I just created a market and lost all the context.

To add on top of this:

After creating a market how can I check the markets I created? 
There should be an easy way for this (same for the markets I added liquidity), we need.
- My questions
- My LP positions
(maybe this should be inside 'my bets", although the name might need to change.

Also, how can I create a market? What's the current decision, to omit this for now?

I don't mind being out of scope of this PR, but we need to add a task to do this next.

Diogomartf:
image

what is `templated_id`?

app/components/ui/SwapInput.tsx
berteotti: We should handle this in a new task. Will delete the comment and create a task.

hooks/contracts/kleros.ts
berteotti: Since it's something contract related, I fill like it's on a good spot. It's related to contracts and it's related to kleros, so `contracts/kleros` makes sense for me. Having a bunch of different data inside the same constants folder, not sure if it's the best approach, since it's difficult to know what constants are inside that file. This way I can go through the folder structure and easily find data related to kleros contracts. What do you think about this?

berteotti:
> Agree on this, I just created a market and lost all the context.

We can redirect the user to the created market. What do you think?

> * My questions
> * My LP positions

I think we need a path to see created markets, maybe in the same page you can see created markets and LP'd ones.

> Also, how can I create a market? What's the current decision, to omit this for now?

Currently there is no button on the navbar, I think we need to figure out how to organize the navbar to fit all the links. I would hide it for now and we can add it later.

Diogomartf:
> > Agree on this, I just created a market and lost all the context.
> 
> We can redirect the user to the created market. What do you think?
> 
> > * My questions
> > * My LP positions
> 
> I think we need a path to see created markets, maybe in the same page you can see created markets and LP'd ones.
> 
> > Also, how can I create a market? What's the current decision, to omit this for now?
> 
> Currently there is no button on the navbar, I think we need to figure out how to organize the navbar to fit all the links. I would hide it for now and we can add it later.

I think redirect to the market works.

I'm okay with not having the lists for my LP, and created markets but would be nice to create a task for that.

hooks/contracts/kleros.ts
Diogomartf: it seems a bit confusing, but will let you decide, we can always change later if doesn't stick.

But why inside hooks folder?

berteotti: Because it's contract related stuff. But will move it from here

PR #150, opened 2024-12-27
chore(PGIO-119): Create function to return explorer link

ElRodrigote:
## Fixes: [PGIO-119](https://linear.app/swaprhq/issue/PGIO-119/create-function-to-return-explorer-link)

## Description
* Creates a utils function to get the explorer URL for TXs and addresses
* Cleans the now unneeded old functions

### DIFF
--- a/utils/token.ts
+++ b/utils/token.ts
@@ -1,7 +1,11 @@
 import { GNOSIS_SCAN_URL } from '@/constants';
 
-export const getExplorerUrl = (txHash: string, isTx: boolean = true) => {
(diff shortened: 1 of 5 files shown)


utils/token.ts
```diff
@@ -1,7 +1,8 @@
 import { GNOSIS_SCAN_URL } from '@/constants';
 
+export const getExplorerUrl = (txHash: string, isTx: boolean = true) => {
```
berteotti, line 3: Instead of using a boolean, lets have 2 fuctions, one for transactions and other for addresses. That way we know what it does by the name.

berteotti (approved):
Looking good!

PR #151, opened 2025-01-06
feat(PGIO-131): add ranking algorithm to landing page

ElRodrigote:
## Fixes: [PGIO-131](https://linear.app/swaprhq/issue/PGIO-131/add-ranking-algorithm-to-landing-page)

## Description
* Adds liquidity filters
* Sets "Most Active" as the default filter
* Adds secondary sorting logic

### DIFF
--- a/queries/omen/markets.ts
+++ b/queries/omen/markets.ts
@@ -1,6 +1,7 @@
 import { gql, request } from 'graphql-request';
 import {
   FixedProductMarketMaker_Filter,
+  FixedProductMarketMaker_OrderBy,
   FpmmTrade_Filter,
   FpmmTransaction_Filter,
   Query,
@@ -126,8 +127,9 @@ const marketDataFragment = gql`
     }
     outcomes
     outcomeTokenMarginalPrices
-    usdVolume
     resolutionTimestamp
+    usdRunningDailyVolume
+    usdVolume
     currentAnswer
     currentAnswerTimestamp
     fee
@@ -406,6 +408,12 @@ const getMarket = async (params: QueryFixedProductMarketMakerArgs) => {
   return !hasDangerousKeywords ? response : null;
 };
 
+type PrimaryOrderBy =
+  | FixedProductMarketMaker_OrderBy.UsdRunningDailyVolume
+  | FixedProductMarketMaker_OrderBy.CreationTimestamp
+  | FixedProductMarketMaker_OrderBy.UsdVolume
+  | FixedProductMarketMaker_OrderBy.ScaledCollateralVolume
+  | FixedProductMarketMaker_OrderBy.OpeningTimestamp;
 const getMarkets = async (
   params: QueryFixedProductMarketMakersArgs & FixedProductMarketMaker_Filter
 ) => {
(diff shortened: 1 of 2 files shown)


queries/omen/markets.ts
```diff
@@ -406,6 +408,12 @@ const getMarket = async (params: QueryFixedProductMarketMakerArgs) => {
   return !hasDangerousKeywords ? response : null;
 };
 
+type PrimaryOrderBy =
+  | FixedProductMarketMaker_OrderBy.UsdRunningDailyVolume
+  | FixedProductMarketMaker_OrderBy.CreationTimestamp
```
Diogomartf, line 413: I think we should value more the remaining days, the less the better, then the creation timestamp

queries/omen/markets.ts
```diff
@@ -419,7 +427,22 @@ const getMarkets = async (
     fpmm => !marketHasDangerousKeyword(fpmm)
   );
 
-  return { fixedProductMarketMakers: filteredResults };
+  const sortedResults = filteredResults.sort((a, b) => {
```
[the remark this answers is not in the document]
ElRodrigote: I think it's nice to have this feature, applying a secondary sort criterion if we get a primary sort result that returns many markets with the same sorted value (for example, the same volume or same liq).

AFAIK the Graph doesn't allow multiple values for the `orderBy` query parameter.

This function will only sort a given page results (as we also apply a `first` and `skip` parameter values to the query result)

queries/omen/markets.ts
```diff
@@ -419,7 +427,22 @@ const getMarkets = async (
     fpmm => !marketHasDangerousKeyword(fpmm)
   );
```
berteotti, line 429: ` onst orderBy = params.orderBy as PrimaryOrderBy;`

Let's to something like this to make it easier to read.

queries/omen/markets.ts
berteotti, line 430: Let's leave a comment here so we can understand why we are doing this.

queries/omen/markets.ts
```diff
@@ -419,7 +427,30 @@ const getMarkets = async (
...
-  return { fixedProductMarketMakers: filteredResults };
+  /**
+   * The Graph doesn't allow multiple values in the orderBy parameter (at least for now).
+   *
+   * This fn applies a secondary sort criterion if we get a primary sort result that
+   * has multiple markets with the same sorted value (for example, the same volume
+   * or same liq).
+   * This fn ONLY SORTS a given page results. (as we also apply a first and skip parameter
+   * values to the query result)
+   */
```
berteotti, line 438:
```suggestion
  /**
   * The Graph doesn't allow to order by multiple fields.
   *
   * This fn applies a secondary sort criteria if we get a primary sort result that
   * has multiple markets with the same sorted value.
   * This fn only sorts a given page results. 
   */
```

What do you think about this? To be simpler

ElRodrigote: Awesome! Thanks

berteotti (approved):
Looks good to me. Let's wait for @Diogomartf review

Diogomartf (approved):

PR #152, opened 2025-01-07
feat(PGIO-136): use agent label instead of address

ElRodrigote:
## Fixes: [PGIO-136](https://linear.app/swaprhq/issue/PGIO-136/use-agent-label-instead-of-address)

## Description
* Displays the AI agent label instead of its address if able

## Visual Evidence
https://www.loom.com/share/2f1ab205f0234d28994293d21bd80e9d?sid=968b6f97-93d6-4b0b-a23f-65f9ca7ce2b2


app/(main)/markets/MarketActivity.tsx
```diff
@@ -93,15 +96,15 @@ export const MarketActivity = ({ id }: { id: string }) => {
...
+              const aiAgent = getIsAIAgent(transaction.user.id);
               const isLiquidityEvent = transaction.fpmmType === FpmmType.Liquidity;
 
               if (isLiquidityEvent)
                 return (
                   <LiquidityEventRow
                     key={transaction.id}
                     transaction={transaction}
-                    isAIAgent={isAIAgent}
+                    aiAgent={aiAgent as AiAgent | undefined}
```
berteotti, line 107: Let's return the type on the `getIsAIAgent` function.

app/(main)/profile/page.tsx
```diff
@@ -267,7 +276,7 @@ export default function ProfilePage() {
             <AddressLink
               href={getExplorerAddressUrl(address)}
               address={address}
-              isAIAgent={isAIAgent}
+              aiAgent={aiAgent as AiAgent | undefined}
```
berteotti, line 279: Let's return the type on the `getIsAIAgent` function.

app/types/index.ts
berteotti, line 1: Let's have this file outside the app dir.

app/(main)/profile/page.tsx
ElRodrigote: I tried to do that but the return type on that `useMemo` fn is different, it returns:
`Record<string, unknown> | undefined`

But I don't know why it returns a record with `unknown` in that type, that's why I casted it.

app/types/index.ts
ElRodrigote: You're right, I thought I've placed it at root level, thanks !!

berteotti (approved):
Looking good!

PR #153, opened 2025-01-15
feat(PGIO-137): add pagination to My Bets

ElRodrigote:
## Fixes: [PGIO-137](https://linear.app/swaprhq/issue/PGIO-137/add-pagination-to-my-bets)

## Description
* Updates the user position query to allow for `skip` and `first` parameters (needed to emulate pagination)
* Creates the new `<Pagination />` component, implements the new component where needed
* Updates the Profile and My Bets pages to allow pagination when needed. Creates support for search params pagination.

## Visual Evidence
https://www.loom.com/share/c2034270a73945299d6129d711c513af?sid=e8f0bafe-7e3f-4fd6-b0cb-1495ec27b871


app/(main)/markets/MarketActivity.tsx
```diff
@@ -130,25 +129,7 @@ export const MarketActivity = ({ id }: { id: string }) => {
...
-          <Button
-            variant="pastel"
-            onClick={() => setPage(page + 1)}
-            className={cx('space-x-2', { invisible: !hasMoreMarkets })}
-          >
-            <p>Next</p>
-            <Icon name="chevron-right" />
-          </Button>
-        </div>
+        <Pagination isFinalPage={!hasMoreMarkets} page={page} setPage={setPage} />
```
Diogomartf, line 132: let's move the pagination bellow

Diogomartf:
Design wise, let's move the pagination bellow, it's cleaner. Even tho I know it can be a lot of elements, will probably add search bet on that page at some point in the future.

Diogomartf:
Let's remove the BG color

image

app/components/Pagination.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Button, Icon } from '@swapr/ui';
+
+interface PaginationProps {
+  isFinalPage: boolean;
+  page: number;
+  setPage: (page: number) => void;
+}
+
+export const Pagination = ({ isFinalPage, page, setPage }: PaginationProps) => (
+  <div className="flex w-full items-center justify-between bg-surface-surface-1 p-4">
```
Diogomartf, line 14: let's remove the bg color 

```suggestion
  <div className="flex w-full items-center justify-between p-4">
```

app/components/BetsListPanel.tsx
```diff
@@ -55,6 +62,9 @@ export const BetsListPanel = ({
         bets.map((userBet: UserBet) => (
           <CardComponent userBet={userBet} key={userBet.id} />
         ))}
+      {page && (page !== 1 || hasNextPage) && setPage && (
+        <Pagination isFinalPage={!hasNextPage} page={page} setPage={setPage} />
+      )}
```
Diogomartf, line 67: we can embed this logic inside of the component since it has all the props to make it happen, or I'm missing something?

app/(main)/profile/page.tsx
```diff
@@ -60,8 +74,20 @@ export default function ProfilePage() {
...
+  });
+
+  const { data: nextPagePositions } = useQuery<UserBet[]>({
+    queryKey: ['getUserBetsNextPage', address, page],
+    queryFn: async () =>
+      await getUserBets({
+        address,
+        itemsPerPage: ITEMS_PER_PAGE,
+        page: page + 1,
+      }),
```
Diogomartf, line 90: I know we've done it like this in past, but would be more efficient to query how many items does it have?

Like doing a simple query with the max, 999, as see how much items it has. Most of the queries will be bellow that, and we have `totalItems`, in case it's more we can do another query when changing page.

Diogomartf: we can address this in another PR, I think we need to revaluate how we do pagination on thegraph

Diogomartf:
The pagination from history seems very similar to yours (design wise), maybe we can use same component?

image

app/components/SimplePagination.tsx
```diff
@@ -0,0 +1,38 @@
...
+  page?: number;
+  setPage?: (page: number) => void;
+}
+
+export const SimplePagination = ({
+  isFinalPage,
+  page,
+  setPage,
+}: SimplePaginationProps) =>
+  page && (page !== 1 || !isFinalPage) && setPage ? (
```
Diogomartf, line 18: instead of ternary can we have an early return? 

I think it's cleaner to read

`!(page && (page !== 1 || !isFinalPage) && setPage) return null`

ElRodrigote:
> The pagination from history seems very similar to yours (design wise), maybe we can use same component?
> 
> image

Just in case, if you mean the MarketActivity tab in the Market Details history, I already replaced the old component with the new SimplePagination one :D

app/(main)/profile/page.tsx
```diff
@@ -37,13 +37,27 @@ import { useEnsName } from 'wagmi';
 import { mainnetConfigForENS } from '@/providers/chain-config';
 import { mainnet } from 'viem/chains';
 
-import { AiAgent } from '@/types';
-
+const ITEMS_PER_PAGE = 100;
```
berteotti, line 40: Why 100 items? Can we show 10 per page? I see that all bets depends on this number, maybe we can fix the "all bets number problem" in another PR

app/(main)/my-bets/page.tsx
```diff
@@ -7,17 +7,41 @@ import { UserBet, UserBetsManager } from '@/entities';
...
 import { useQuery } from '@tanstack/react-query';
 
-import { useMemo } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { TabBody, TabGroup, TabHeader } from '@swapr/ui';
 import { useAccount } from 'wagmi';
 import { BetsListPanel, BetsListTab } from '@/app/components';
+import { useRouter } from 'next/navigation';
+
+const ITEMS_PER_PAGE = 50;
```
berteotti, line 16: Can we show 10 items per page?

app/(main)/profile/page.tsx
ElRodrigote: Yeah, all the filters depends on the "All Bets", and the "All Bets" is the result from getting the query result from The Graph and then applying a filter in the FE, so our simulated pagination sometimes behave in an odd way.

Anyway, the other tabs are filters, so they'll never be `> All bets`

berteotti:
Something isn't working properly with page size.

10 page size:
image

25 page size:
image

PR #154, opened 2025-01-16
Update Profile Stats with Dune Data

Diogomartf:
image

**summary** 
- Get data from dune query that gets trade metrics for all addresses on omen
- add ability to filter by address
- connect component with new data
- remove unused component

**Fixes**
- accuracy (previously with the graph only profiles that had less 999 items were being accounted), so bigger tradding accounts had wrong data.


app/(main)/profile/page.tsx
```diff
@@ -31,19 +31,29 @@ import {
   ProfileCardBet,
   StatsCard,
 } from '@/app/components';
-import { getAIAgents } from '@/queries/dune';
+import {
+  getAIAgents,
+  getAgentsLeaderboardData,
```
ElRodrigote, line 36: We can remove this if we're not using it.

queries/dune/queries.ts
```diff
@@ -1,4 +1,9 @@
-import { LatestResultArgs, ParameterType, RunQueryArgs } from '@duneanalytics/client-sdk';
+import {
+  LatestResultArgs,
+  ParameterType,
+  QueryParameter,
```
ElRodrigote, line 4: We can remove this if we're not using it.

queries/dune/queries.ts
```diff
@@ -100,3 +107,39 @@ export const getAgentsTotalsData = async () => {
...
+export const getOmenTradeMetrics = async ({
+  page = 1,
+  pageSize = 50,
+  sort_by = 'profit_loss desc',
+  filters = '',
+}) => {
+  const limit = pageSize;
+  const offset = (page - 1) * pageSize;
+  const filtersString = filters ? `&filters=${filters}` : '';
+  console.log('filtersString:', filtersString);
```
ElRodrigote, line 122:
```suggestion
```

queries/dune/queries.ts
```diff
@@ -100,3 +107,39 @@ export const getAgentsTotalsData = async () => {
...
+  const url = `https://api.dune.com/api/v1/query/${DUNE_OMEN_TRADE_METRICS}/results?limit=${limit}&offset=${offset}&sort_by=${sort_by}${filtersString}`;
```
ElRodrigote, line 123: Maybe we can abstract this into a utils fn, WDYT?

ElRodrigote (approved):

queries/dune/queries.ts
Diogomartf: I think url is okay here for now, I can put it in a variable, but since it's only used here I'd rather have it in this file.

ElRodrigote: Yes!

I meant to create a util fn in this file to take the query ID, limit, offset, sortby and filters string. No prob!

Diogomartf: Yes, did some abstractions since the code was being reused.

ElRodrigote (approved):

app/(main)/profile/page.tsx
```diff
@@ -296,26 +190,32 @@ export default function ProfilePage() {
...
           symbol="usd"
-          isLoading={isLoadingStats}
+          isLoading={isLoadingTradeMetrics}
         />
         <StatsCard
           title="Success rate"
-          value={successRate}
+          value={tradeMetrics?.success_rate?.toString()}
           symbol="%"
           isLoading={isLoadingStats}
```
berteotti, line 214:
```suggestion
          isLoading={isLoadingTradeMetrics}
```

berteotti (approved):
Looking good!

PR #155, opened 2025-01-17
feat(PGIO-143): remove dangerous keywords filter

ElRodrigote:
## Fixes: [PGIO-143](https://linear.app/swaprhq/issue/PGIO-143/remove-dangerous-keywords-filter)

## Description
* Creates a new disclaimer component
* Removes an unwanted markets filter

## Visual Evidence
Screenshot 2025-01-21 at 1 02 58 PM

### DIFF
--- a/app/components/DangerousWordInfo.tsx
+++ b/app/components/DangerousWordInfo.tsx
@@ -1,26 +1,12 @@
 'use client';
 
-import { cx } from 'class-variance-authority';
-
 import { Icon, Tooltip, TooltipContent, TooltipTrigger } from '@swapr/ui';
 
-interface DangerousWordInfoProps {
-  small?: boolean;
-}
-
-export const DangerousWordInfo = ({ small }: DangerousWordInfoProps) => (
+export const DangerousWordInfo = () => (
   <Tooltip>
     <TooltipTrigger>
-      <div
-        className={cx(
-          'flex items-center gap-1 rounded-12 border-2 border-text-danger-main bg-text-danger-em px-2 py-1.5 text-sm font-semibold text-outline-warning-base-em',
-          {
-            'rounded-6': small,
-          }
-        )}
-      >
+      <div className="flex items-center gap-1 rounded-6 border-2 border-text-danger-main bg-text-danger-em px-2 py-1.5 text-sm font-semibold text-outline-warning-base-em">
         <Icon name="warning" size={16} />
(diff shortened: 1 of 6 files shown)


berteotti:
I think only this one is enough for now. This will simplify the too code.

image

app/components/DangerousWordInfo.tsx
```diff
@@ -0,0 +1,33 @@
...
+            'rounded-6': small,
+          }
+        )}
+      >
+        <Icon name="warning" size={16} />
+        {!small && <p>Potentially inappropriate</p>}
+      </div>
+    </TooltipTrigger>
+    <TooltipContent className="bg-surface-surface-0">
+      <p>This market name contains words that may be considered inappropriate.</p>
```
berteotti, line 27:
```suggestion
      <p>This market may be inappropriate to bet on.</p>
```

I think something around this works might be better

ElRodrigote: Done, thanks!

berteotti (approved):

Diogomartf:
Here we should use yellow color, like `surface-warning-main`. As it's a warning.

image

Red should be reserved for urgent actions.

Eg:

image

or something more subtle, also works.

image

app/components/DangerousWordInfo.tsx
```diff
@@ -0,0 +1,19 @@
+'use client';
+
+import { Icon, Tooltip, TooltipContent, TooltipTrigger } from '@swapr/ui';
+
+export const DangerousWordInfo = () => (
+  <Tooltip>
+    <TooltipTrigger>
+      <div className="flex items-center gap-1 rounded-6 border-2 border-text-danger-main bg-text-danger-em px-2 py-1.5 text-sm font-semibold text-outline-warning-base-em">
```
Diogomartf, line 8: let's remove red and use yellow

app/components/DangerousWordInfo.tsx
```diff
@@ -0,0 +1,19 @@
+'use client';
+
+import { Icon, Tooltip, TooltipContent, TooltipTrigger } from '@swapr/ui';
+
+export const DangerousWordInfo = () => (
```
Diogomartf, line 5: I think, `DangerousWordsWarning`, describes better

```suggestion
export const DangerousWordsWarning = () => (
```

Diogomartf (approved):

berteotti (approved):

PR #162, opened 2025-02-03
feat: update market insights endpoint

berteotti:
- Market insights endpoint will accept questions instead of market id


app/(main)/markets/News.tsx
```diff
@@ -6,9 +6,10 @@ import Image from 'next/image';
 import { FA_EVENTS } from '@/analytics';
 import { Skeleton } from '@/app/components';
 import { getMarketInsights, MarketInsights } from '@/queries/gnosis-ai';
+import { FixedProductMarketMaker } from '../../../queries/omen';
```
Diogomartf, line 9: can we use @ syntax? 

```suggestion
import { FixedProductMarketMaker } from '@/queries/omen';
```

PR #163, opened 2025-02-03
Improve loading with loading dots and some UI improvements

Diogomartf:
- adds Loading effect with `LoadingDots` to loading message
- renames `MARKET_CHAT_URL` to `PRESAGIO_CHAT_API_URL`
- remove blur from bottom chat box as it's not working

**screenshots**
image


app/components/MessageGradientTextLoading.tsx
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export const MessageGradientTextLoading: React.FC<MessageGradientTextLoadingProps> = ({
+  children,
+  className = '',
+}) => {
+  return (
+    <div className="rounded-20 bg-outline-primary-base-em px-4 py-2">
+      <div className={className}>
+        <span className="animate-gradient-x to- to- bg-gradient-to-r from-text-low-em via-text-high-em bg-clip-text text-transparent">
```
berteotti, line 14: What about animate `animate-pulse` here?

app/components/MessageGradientTextLoading.tsx
```diff
@@ -0,0 +1,20 @@
+import { PropsWithChildren } from 'react';
+
+interface MessageGradientTextLoadingProps extends PropsWithChildren {
+  className?: string;
+}
+
+export const MessageGradientTextLoading: React.FC<MessageGradientTextLoadingProps> = ({
```
berteotti, line 7: Why a new component just for elements with style?

Diogomartf: we might need it in more places, but yeah, we can do it inline right now.

berteotti (approved):

PR #166, opened 2025-02-10
Add search filters for address/label on Dune query

Diogomartf:
Adds search filter support for address and labels to agents leaderboard

image
image

### DIFF
--- a/app/(main)/leaderboard/agents/AgentsLeaderboardTable.tsx
+++ b/app/(main)/leaderboard/agents/AgentsLeaderboardTable.tsx
@@ -11,13 +11,14 @@ import {
   TableHeader,
   TableRow,
 } from '@/app/components/ui/Table';
-import { Button, Icon, IconButton } from '@swapr/ui';
+import { Button, Icon, IconButton, Input } from '@swapr/ui';
 import { Avatar, Skeleton } from '@/app/components';
 import { useQuery } from '@tanstack/react-query';
 import { getAgentsTradeMetricsData } from '@/queries/dune';
 import { twMerge } from 'tailwind-merge';
 import Link from 'next/link';
 import { formatValueWithFixedDecimals } from '@/utils';
+import { useDebounce } from '@/hooks';
 
 const ITEMS_PER_PAGE = 100;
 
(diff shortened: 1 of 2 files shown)


app/(main)/leaderboard/agents/AgentsLeaderboardTable.tsx
```diff
@@ -47,18 +48,31 @@ export default function AgentsLeaderboardTable() {
...
+      sortKey,
+      sortOrder,
+      filter,
+    ],
     queryFn: () =>
       getAgentsTradeMetricsData({
         page,
         pageSize: ITEMS_PER_PAGE,
         sort_by: `${sortKey} ${sortOrder}`,
+        filters: filter ? `address = ${filter} or label = ${filter}` : '',
```
ElRodrigote, line 75: Is there some sort of `like` or `like_operator` we could use here? Right now it works for exact matches (at least what I could test so far) but it would be awesome searching for "olas" and getting all the names that contains `olas` (upper case, lower case, etc)

ElRodrigote (approved):

app/(main)/leaderboard/agents/AgentsLeaderboardTable.tsx
```diff
@@ -114,115 +140,133 @@ export default function AgentsLeaderboardTable() {
...
-                )}
+      <Input
+        leftIcon="search"
+        className="mb-4 w-full"
+        placeholder="Search agent by address or label"
+        value={inputValue}
+        onChange={e => setInputValue(e.target.value)}
+      />
+      {isLoading ? (
+        <LoadingLeaderBoardTable />
```
berteotti, line 163: I think this loading is too big and the page jumps a lot when you do a search, maybe we can have a smaller loading?

app/(main)/leaderboard/agents/AgentsLeaderboardTable.tsx
Diogomartf: yes, it would be awesome, I'm trying to see how can we do that since it's not clear on Dune docs, but I'm their discord trying to figure it out.

app/(main)/leaderboard/agents/AgentsLeaderboardTable.tsx
Diogomartf: I reduced the number of rows on the LoadingTable, it's gets a bit better, but it doesn't solve the problem.

app/(main)/leaderboard/agents/AgentsLeaderboardTable.tsx
Diogomartf: done!

berteotti (approved):

PR #172, opened 2025-02-19
Use the MarketsHighlight component to highlight 3 markets and make landing more interesting

Diogomartf:
Since we already have some ranking and the first 12 markets should be the best in terms of liquidity and betting activity, we can randomly highlight the markets, since it makes the page more interesting.

**preview**

image

### DIFF
--- a/app/(main)/MarketsHighlight.tsx
+++ b/app/(main)/MarketsHighlight.tsx
@@ -17,7 +17,7 @@ import { FA_EVENTS } from '@/analytics';
 import { trackEvent } from 'fathom-client';
 
 interface MarketsHighlightProps {
-  markets: any[];
+  markets: FixedProductMarketMaker[];
 }
 
 export const MarketsHighlight = ({ markets }: MarketsHighlightProps) => {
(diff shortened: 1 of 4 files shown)


app/(main)/MarketsHighlight.tsx
```diff
@@ -14,28 +13,15 @@ import {
...
-      return {
-        data: results
-          .map(result => result.data?.fixedProductMarketMaker)
-          .filter((market): market is FixedProductMarketMaker => !!market),
-        isLoading: results.some(result => result.isPending),
-      };
-    },
-  });
+interface MarketsHighlightProps {
+  markets: any[];
```
ElRodrigote, line 20: Is there a way we could type this or it gets too complex?

ElRodrigote (approved):

app/(main)/MarketsHighlight.tsx
berteotti: We can use `FixedProductMarketMaker[]` here

Diogomartf: fixed

PR #175, opened 2025-02-24
truncate label on MarketActivity rows

Diogomartf:
This update makes all the table columns visible without having to do horizontal scroll.

**now**
image

We're still able to see the full name on hover
image

**previous**
image


app/components/LabelWithLink.tsx
```diff
@@ -12,22 +12,22 @@ import { useEnsName } from 'wagmi';
...
+interface LabelWithLinkProps extends ComponentProps<typeof Link> {
   address: Address;
   aiAgent?: AiAgent;
   className?: string;
   href: string;
   iconSize?: number;
 }
 
-export const AddressLink = ({
+export const LabelWithLink = ({
```
berteotti, line 23: Why not remove With? I think it's a label and a link

berteotti (approved):

app/components/LabelWithLink.tsx
Diogomartf: it's a bit of logic to handle the "ai" label, we used it in 2 places, so let's keep for now.

PR #179, opened 2025-03-25
Update swapr UI theme 1.0

Diogomartf:
The PR is meant to update the styles to new theme, and use updated components

## screenshots

image
image
image
image
image
image

image
image
image
image
image
image
image
image
image


ElRodrigote (approved):

app/components/Navbar.tsx
```diff
@@ -15,9 +15,7 @@ import { useMemo } from 'react';
 
 import { NetworkButton } from './NetworkButton';
 import { trackEvent } from 'fathom-client';
-import BetsIcon from '@/app/components/ui/icons/BetsIcon';
-import AIStars from '@/public/ai-stars.svg';
-import Image from 'next/image';
```
ElRodrigote, line 20: We removed these 2 but we're not removing the `<Link />` component that implements the button and image depending on these 2 imports.

Do we still need these imports?
Should we remove the implementation at line 58?

app/components/SettingsPopover.tsx
```diff
@@ -33,15 +33,20 @@ export const SettingsPopover = () => {
...
+        />
       </PopoverTrigger>
       <PopoverContent className="max-w-lg px-4">
         <div className={address && 'space-y-3 divide-y divide-outline-base-em'}>
           {address && (
             <div className="space-y-2">
               <ButtonLink
                 size="sm"
-                variant="pastel"
+                variant="secondary"
```
ElRodrigote, line 49: At lines 63 and 74 I think we should do the same change we did here, these lines still keep the `variant="pastel"`

app/components/Navbar.tsx
```diff
@@ -15,9 +15,7 @@ import { useMemo } from 'react';
 
 import { NetworkButton } from './NetworkButton';
 import { trackEvent } from 'fathom-client';
-import BetsIcon from '@/app/components/ui/icons/BetsIcon';
```
ElRodrigote, line 18: If we don't use the `Icon` and `LifiWidgetPopover` components, we could remove those too.

ElRodrigote (changes requested):
The build is breaking because we are implementing deleted components in the Navbar, left some comments about that.

OPTIONAL: suggested changes but not breaking the build
We still have `variant="pastel"` in:
* `app/(main)/chat/Chat.tsx`
* `app/(main)/chat/new/page.tsx`

We still have `variant="outline"` in:
* `app/(main)/chat/ChatSidebar.tsx`

We still have `variant="solid"` in:
* `app/(main)/chat/SessionScreen.tsx`
* `app/components/AiChatBase.tsx`

app/components/SettingsPopover.tsx
Diogomartf: I think something was off on your diff, cause it's updated

app/components/Navbar.tsx
Diogomartf: We can remove Icon, `LifiWidgetPopover` was removed by the merged now fixed. thanks

Diogomartf:
> The build is breaking because we are implementing deleted components in the Navbar, left some comments about that.
> 
> OPTIONAL: suggested changes but not breaking the build We still have `variant="pastel"` in:
> 
> * `app/(main)/chat/Chat.tsx`
> * `app/(main)/chat/new/page.tsx`
> 
> We still have `variant="outline"` in:
> 
> * `app/(main)/chat/ChatSidebar.tsx`
> 
> We still have `variant="solid"` in:
> 
> * `app/(main)/chat/SessionScreen.tsx`
> * `app/components/AiChatBase.tsx`

forgot to push the update 🫣 Sorry, updated now.

ElRodrigote (approved):

berteotti (approved):
Awesome job, looks awesome the new design

PR #180, opened 2025-04-08
feat: add analytics to chat and improve ux

berteotti:
- Adds analytics to chat
- Refreshes sidenav list after creating new chat

### DIFF
--- a/analytics/index.ts
+++ b/analytics/index.ts
@@ -29,16 +29,16 @@ export const FA_EVENTS = {
     ROUTE_SUCCESS: 'click/lifi-widget/route-success',
   },
   AI_CHAT: {
-    LANDING: {
-      OPEN: 'click/ai-chat/landing/open',
+    WIZARD: {
+      OPEN: 'click/ai-chat/wizard/open',
+      OPEN_LANDING: 'click/ai-chat/wizard/open/landing',
+      OPEN_MARKET: (id: string) => `click/ai-chat/wizard/open/${id}`,
     },
-  },
-  MARKET: {
-    AI_CHAT: {
-      OPEN_GENEREAL: 'click/market/ai-chat/open-general',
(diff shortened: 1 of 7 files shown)


analytics/index.ts
```diff
@@ -4,6 +4,13 @@ const formatEventName = (eventName: string) =>
   eventName.split(' ').join('-').toLocaleLowerCase();
 
 export const FA_EVENTS = {
+  CHAT: {
+    NEW: 'click/chat/new',
+    NAVBAR: 'click/chat/navbar',
+    WIZARD: 'click/chat/wizard',
+    OPEN: (id: string) => `click/chat/${id}`,
+    OPEN_WIZARD: (id: string) => `click/chat/wizard/${id}`,
```
Diogomartf, line 12: should we track when there is multiple messages or we check that on the db?

ElRodrigote, line 12: Aren't these events colliding with the events at line 44?

I feel we're double-tracking some of the `OPEN` events, but I'm not sure.

[In this file](https://github.com/SwaprHQ/presagio/blob/8f2b38b8abb5cb9f6de9e980926a61ab96994142/app/components/AiChatBase.tsx#L148), we're tracking the chat opening when we click the Wizard icon, but we're also tracking the chat start now when we click the button.

ElRodrigote: I don't know what the goal is on tracking if there were multiple messages in the chat. 

Maybe we could check if there is one or more messages from the user to know if the users are chatting with the bot?

berteotti: Nice catch, I will merge with the other events. Ones should point if the wizard was opened, others should point if the user created a new chat.

Diogomartf (approved):

ElRodrigote (approved):

PR #185, opened 2025-04-14
add sentry for error tracking

Diogomartf:
image

**link**
https://swaprhq.sentry.io/projects/presagio-app/?project=4509152954875984

**Also in this PR**
- update next to 15.3.0
- transform next config file into ts


instrumentation-client.ts
```diff
@@ -0,0 +1,13 @@
+import * as Sentry from '@sentry/nextjs';
+
+Sentry.init({
+  dsn: 'https://36be948fdd4ed19a7814987b57b6da58@o4509152952254464.ingest.de.sentry.io/4509152954875984',
```
berteotti, line 4: Should this be on an env variable?

sentry.edge.config.ts
```diff
@@ -0,0 +1,16 @@
+// This file configures the initialization of Sentry for edge features (middleware, edge routes, and so on).
+// The config you add here will be used whenever one of the edge features is loaded.
+// Note that this config is unrelated to the Vercel Edge Runtime and is also required when running locally.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from '@sentry/nextjs';
+
+Sentry.init({
+  dsn: 'https://36be948fdd4ed19a7814987b57b6da58@o4509152952254464.ingest.de.sentry.io/4509152954875984',
```
berteotti, line 9: Maybe we should have this in env var since it's the same in 3 places

instrumentation-client.ts
Diogomartf: I think it's okay, had the same thought but that was default from sentry codemon, I think it's simpler for now to have it directly.

berteotti (approved):

PR #187, opened 2025-04-17
Fix highlighted markets changing when filters applied

Diogomartf:


app/(main)/MarketsHighlight.tsx
```diff
@@ -15,13 +15,17 @@ import {
...
 interface MarketsHighlightProps {
   markets: FixedProductMarketMaker[];
 }
 
 export const MarketsHighlight = ({ markets }: MarketsHighlightProps) => {
-  const randomMarkets = markets.sort(() => 0.5 - Math.random()).slice(0, 3);
+  const randomMarkets = useMemo(
+    () => markets.sort(() => 0.5 - Math.random()).slice(0, 3),
+    [markets]
+  );
```
ElRodrigote, line 28: MINOR, non-blocking

Depending on the markets array length this shuffling might get biased and/or expensive for the client, but the code is easy to read.

If you think it's worth it, there's an alternative called Fisher-Yates for shuffling finite lists, it's a bit more verbose but is faster and cheaper (and it shuffles the array with less bias)

```tsx
  // Fisher-Yates shuffle for better randomization
  const shuffledMarkets = useMemo(() => {
    const randomized = [...markets]; // -> avoid mutating the original
    
    // this `for()` block is the Fisher-Yates algorithm
    for (let i = randomized.length - 1; i > 0; i--) {
      const j = Math.floor(Math.random() * (i + 1));
      [randomized[i], randomized[j]] = [randomized[j], randomized[i]];
    }
    
    return randomized.slice(0, 3);
  }, [markets]);
```

ElRodrigote (approved):

app/(main)/MarketsHighlight.tsx
Diogomartf: Looks cool, will add it.

berteotti (approved):

